PR #21, opened 2025-05-12
Llama-stack API key in configuration

tisnik:
## Description

Llama-stack API key in configuration

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update
- [ ] CI configuration change

### DIFF
--- a/lightspeed-stack.yaml
+++ b/lightspeed-stack.yaml
@@ -1,1 +1,4 @@
 name: foo bar baz
+llama_stack:
+  url: http://localhost:8321
+  api_key: xyzzy
--- a/src/models/config.py
+++ b/src/models/config.py
@@ -1,12 +1,13 @@
-from typing import Optional
-
 from pydantic import BaseModel
+
+from typing import Optional
 
 
 class LLamaStackConfiguration(BaseModel):
     """Llama stack configuration."""
 
     url: str
+    api_key: Optional[str] = None
 
 
 class Configuration(BaseModel):


lightspeed-stack.yaml
```diff
@@ -1 +1,4 @@
 name: foo bar baz
+llama_stack:
+  url: http://localhost:8321
```
manstis, line 3: Are you planning on running `llama-stack` (server) separately to your own `uvicorn` instance?

The _default_ `server` launched with `llama-stack` is [here](https://github.com/meta-llama/llama-stack/blob/main/llama_stack/distribution/server/server.py#L317-L502).

I suspect you could use Meta's code as a starting point to integrate `llama-stack` itself into your code?

manstis: Hi @tisnik running `llama-stack` as a separate image may be difficult.

`llama-stack` will take some (IDK how much) configuration from `lightspeed-stack`; providers, models etc. Your distribution would then need to effectively manage two containers; one for `lightspeed-stack` and one for `llama-stack`. 

You may want to look at [this](https://llama-stack.readthedocs.io/en/latest/distributions/importing_as_library.html) too.

i.e. `lightspeed-stack` is the _server_ but uses `llama-stack` as a _library_ internally.

tisnik: That one looks reasonable, thank you @manstis !

tisnik: Tested it right now, definitely possible solution, yes.

PR #40, opened 2025-05-20
[WIP] Streaming Chat with Referenced Documents PoC

TamiTakamiya:
## Description

Streaming Chat with Referenced Documents PoC.  Requires llama-stack 0.2.7.

https://github.com/user-attachments/assets/80ec4800-4a4e-4ee6-9834-68dc0ecd346e

## Type of change

- [ ] Refactor
- [X] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] CI configuration change

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/src/app/endpoints/query.py
+++ b/src/app/endpoints/query.py
@@ -8,7 +8,9 @@ from llama_stack_client.lib.agents.agent import Agent
 from llama_stack_client import LlamaStackClient
 from llama_stack_client.types import UserMessage
 
-from fastapi import APIRouter, Request
+from fastapi import APIRouter
+
+from pydantic import BaseModel
 
 from configuration import configuration
 from models.config import LLamaStackConfiguration
@@ -25,9 +27,11 @@ query_response: dict[int | str, dict[str, Any]] = {
     },
 }
 
+class LLMRequest(BaseModel):
+    query: str
 
 @router.post("/query", responses=query_response)
-def info_endpoint_handler(request: Request, query: str) -> QueryResponse:
+def info_endpoint_handler(request: LLMRequest, query: str) -> QueryResponse:
     llama_stack_config = configuration.llama_stack_configuration
     logger.info("LLama stack config: %s", llama_stack_config)
 
--- a/src/app/endpoints/streaming_query.py
+++ b/src/app/endpoints/streaming_query.py
@@ -1,0 +1,276 @@
+"""Handler for REST API call to provide answer to query."""
+import json
+import logging
+import re
+from typing import Any, Optional, Iterator, AsyncGenerator, Mapping
+
+from llama_stack_client.lib.agents.agent import Agent
+from llama_stack_client.lib.agents.event_logger import TurnStreamPrintableEvent, TurnStreamEventPrinter
+from llama_stack_client import LlamaStackClient
+from llama_stack_client.types import UserMessage
+from llama_stack_client.types.shared.interleaved_content_item import TextContentItem
+
+from fastapi import APIRouter
+from fastapi.responses import StreamingResponse
+
+from app.endpoints.query import get_llama_stack_client, LLMRequest
+from configuration import configuration
+
+logger = logging.getLogger("app.endpoints.handlers")
+router = APIRouter(tags=["streaming_query"])
+
+
+query_response: dict[int | str, dict[str, Any]] = {
+    200: {
+        "query": "User query",
+        "answer": "LLM answer",
+    },
+}
+
+
+@router.post("/streaming_query", responses=query_response)
+def info_endpoint_handler(request: LLMRequest) -> StreamingResponse:
+    llama_stack_config = configuration.llama_stack_configuration
+    logger.info("LLama stack config: %s", llama_stack_config)
+
+    client = get_llama_stack_client(llama_stack_config)
+
+    # retrieve list of available models
+    models = client.models.list()
+
+    # select the first LLM
+    llm = next(m for m in models if m.model_type == "llm")
+    model_id = llm.identifier
+
+    logger.info("Model: %s", model_id)
+
+    response = retrieve_response(client, model_id, request.query)
+
+    return StreamingResponse(
+        response_processing_wrapper(
+            request,
+            response,
+        )
+    );
+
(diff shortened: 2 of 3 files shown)


manstis:
@TamiTakamiya From what I understand this effectively adds _post processing_ of the response _turns_ to add formatting and most importantly extract the "Referenced Documents" from `tool_call="query_from_memory"`? Since, as we discussed yesterday, there will be no _runtime_ "lightspeed-stack" component and "lightspeed-stack" will in essence be an "image builder"/configuration tool the code you propose in this PR should probably be elsewhere.

I am wondering whether it's possible to move all of this to a custom `agent` overriding the [`run`](https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/inline/agents/meta_reference/agent_instance.py#L322C15-L374) method (or another, but this seems a plausible candidate?) to handle the post processing?

We'd then configure the `llama-stack` to use our custom external provider:
```
providers: 
  agent: 
    provider_type: inline::our-super-agent
...
```
At best this code would have to live in `ansible-ai-connect-service` but we should favour trying to write our own `agent`.

WDYT?

manstis:
[`retrieve_response`](https://github.com/lightspeed-core/lightspeed-stack/pull/40/files#diff-016656762a3360d37311e32ba140a113ce19f77a1c56611f8059d466a6cd3db8R56-R98) would be in `ansible-ai-connect-service` as it's using `LlamaStackClient` to invoke `llama-stack` etc.

[`response_processing_wrapper`](https://github.com/lightspeed-core/lightspeed-stack/pull/40/files#diff-016656762a3360d37311e32ba140a113ce19f77a1c56611f8059d466a6cd3db8R102-R118) would live in our custom `Agent` but instead of post-processing `LlamaStackClient` objects it would post-process the `llama-stack` server equivalents.

TamiTakamiya:
@manstis Yes, it's a PoC that is intended to show the possibility to implement the road-core/service `/streaming-chat` endpoint using the llama-stack `builtin::rag/knowledge_search`.  Initially I thought opening a PR for ansible-ai-connect-service, but after I saw the  `/query` endpint in lightspeed-stack, I decided to creating a PR here.  This can be anywhere in between llama-stack and the chatbot client.

It is true that the logic is a bit redundant because the post-processing code parses the streaming output using a regex, stores the results, and then the results are sent at the end of streaming.  If we write our own agent or tool, implementations could be simplified.

The `_yield_printable_events` method of the RAGTurnStreamEventPrinter` class was copied from the `TurnStreamEventPrinter` class and contains lots of duplicated lines. I want to reduce the amount of duplicated lines.

manstis:
@TamiTakamiya Was you wanting to keep this open given your PR for `ansible-ai-connect-service`? Perhaps this is now better refactoring to a `README` or something?

src/app/endpoints/query.py
```diff
@@ -26,8 +28,12 @@
 }
 
 
+class LLMRequest(BaseModel):
+    query: str
```
umago, line 32: Let's keep models in the models/ directory. 

Also, there's now a QueryRequest model that I think we should use instead of introducing a new one:

https://github.com/lightspeed-core/lightspeed-stack/blob/040541782804476587c8efdf05ff3a4c9e2a6122/src/models/requests.py#L55

src/app/endpoints/streaming_query.py
```diff
@@ -0,0 +1,302 @@
...
+    logger.info("Model: %s", model_id)
+
+    response = retrieve_response(client, model_id, request.query)
+
+    return StreamingResponse(
+        response_processing_wrapper(
+            request,
+            response,
+        )
+    )
```
umago, line 58: Can we use the QueryResponse for this ? I don't see why introducing a new StreamingResponse model.

src/app/endpoints/streaming_query.py
```diff
@@ -0,0 +1,302 @@
...
+    logger.info("LLama stack config: %s", llama_stack_config)
+
+    client = get_llama_stack_client(llama_stack_config)
+
+    # retrieve list of available models
+    models = client.models.list()
+
+    # select the first LLM
+    llm = next(m for m in models if m.model_type == "llm")
+    model_id = llm.identifier
```
umago, line 47: Let's keep compatibility with the /query endpoint, which supports passing a model/provider in the request

src/app/endpoints/streaming_query.py
```diff
@@ -0,0 +1,302 @@
...
+    vector_db_id = available_vector_dbs[0]
+
+    agent = Agent(
+        client,
+        model=model_id,
+        instructions="""You are a helpful assistant  with access to the following tools. 
+When a tool is required to answer the user's query, respond only with <|tool_call|> 
+followed by a JSON list of tools used. If a tool does not exist in the provided 
+list of tools, notify the user that you do not have the ability to fulfill the request.
+""",
```
umago, line 83: Ditto, keeping compatibility with the /query endpoint which supports passing a system_prompt in the request

src/app/endpoints/streaming_query.py
```diff
@@ -0,0 +1,302 @@
...
+            }
+        ],
+    )
+    session_id = agent.create_session("chat_session")
+    response = agent.create_turn(
+        messages=[UserMessage(role="user", content=prompt)],
+        session_id=session_id,
+    )
+    return response
+    # return str(response.output_message.content)
```
umago, line 106: A lot of this code is duplicated with /query. I think we can refactor these two endpoints and have a common base of code. Otherwise we will endpoint creating a lot of inconsistencies between the two.

umago (changes requested):
Thanks Tami for this change! I would love to see a streaming endpoint here. But, I do think that this change needs some refactor and make both /query and /streaming_query compatible. I see a lot of potential in sharing code between these two endpoints I think we should do it otherwise we risk making them incompatible in the future.

TamiTakamiya:
@umago @manstis `/streaming_query` endpoint was implemented with https://github.com/lightspeed-core/lightspeed-stack/pull/126.  I will close this and open a new PR only for enabling referenced documents support.

PR #74, opened 2025-06-06
Add /feedback endpoint

umago:
## Description

This patch adds the /feedback endpoint.

Note: Authentication and user ID retrieval haven't been included yet, as they don't appear to be implemented in the Lightspeed Llama-stack version. TODOs were left in the code for those parts.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement

- Related Issue #
- Closes #

## Testing
- Enable the feedback via the configuration file and you can curl it
- Run unit-tests


tisnik:
yup we need this. Will take me some time to review, but I promise to do it

lpiwowar (commented):
I just quickly went through the PR. But overall I think it looks good:)

src/app/endpoints/feedback.py
```diff
@@ -0,0 +1,153 @@
...
+            handle authentication Logic.
+
+    Returns:
+        Response indicating the status of the feedback storage request.
+    """
+    logger.debug("Feedback received %s", str(feedback_request))
+
+    user_id = retrieve_user_id(auth)
+    try:
+        store_feedback(user_id, feedback_request.model_dump(exclude={"model_config"}))
```
lpiwowar, line 101: **thought:** I'm thinking about whether it would make sense to set some upper limit on the length of the user's feedback. This would be to prevent resource exhaustion if some malicious party decides to test out the endpoint.

**question:** Is the plan to store the feedback later somewhere in a DB and not in a plaintext file?

umago: True, we can think of that as well. For now, I'm just mimicking what was present in the previous service.

Regarding the question, looking at the data collection [0], it does seem like it still sends it in plain-text. It's just zipped and sent to console.redhat.com. May be something we want to change too.

But I think these things can be added on top of this work. For now, keeping it compatible to what it was before seems fine IMHO.

[0] https://github.com/road-core/service/tree/643164b47ea1cf9c91bd5a1ce3cadbed1fb3faa4/ols/user_data_collection

lpiwowar: Yes, definitely I think this is a good start! :+1: I was just curious.

Thanks for the link, it makes it a bit clearer to me. I do not know much about the feedback part.

tisnik: basically yes, the plan is to use DB (probably sqlite as it's used a lot in llama-stack itself). The JSON as files solution was easier to implement in openshift scenario, but personally I don't like it very much TBH :)

tisnik: > True, we can think of that as well. For now, I'm just mimicking what was present in the previous service.
> 
> Regarding the question, looking at the data collection [0], it does seem like it still sends it in plain-text. It's just zipped and sent to console.redhat.com. May be something we want to change too.
> 
If the format would change, also the pipelines in RH network will need changes... But doable, of course ;)

lpiwowar: @tisnik I can take a look at the limiting of the feedback. I'll create a ticket:).

lpiwowar: Ticket created:), @umago it is on our board as well. It's a small ticket, but I'm just being transparent. I guess we are still figuring out how to best collaborate together.

> If the format would change, also the pipelines in RH network will need changes... But doable, of course ;)

Yes:), I guess this change would be a bit bigger. I guess if it works, then it is fine for now (?). Maybe something we might want to take later?

PR #159, opened 2025-07-01
LCORE-219: Convert application/yaml as plain/text for attachments/documents

onmete:
## Description

Convert application/yaml as plain/text for attachments/documents

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/src/models/requests.py
+++ b/src/models/requests.py
@@ -79,7 +79,7 @@ class QueryRequest(BaseModel):
     provider: Optional[str] = None
     model: Optional[str] = None
     system_prompt: Optional[str] = None
-    attachments: Optional[list[Attachment]] = None
+    attachments: list[Attachment] = Field(default_factory=list)
 
     # provides examples for /docs endpoint
     model_config = {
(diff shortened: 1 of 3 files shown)


tisnik (changes requested):
that is imho incorrect. We need to pass the right type to LLM in the end

onmete:
@tisnik seems that llama-stack (or something within) doesn't support application/yaml. I've raised a thread in the forum, but I'm not sure if it will be added.

I think this workaround of converting application/yaml to text/plain is semantically correct because:
- YAML is human-readable text - it can be processed as plain text
- Content is preserved - we're only changing the MIME type label, not the actual YAML content
- AI can still understand YAML syntax - whether labeled as application/yaml or text/plain

What about creating some follow-up story to keep an eye on when llama-stack supports application/yaml?

Or do you think there is another route of how to fix it?

manstis:
I'd guess the issue in `llama-stack` is [here](https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/inline/agents/meta_reference/agent_instance.py#L913-L923).

Replacing `application/yaml` with `text/yaml` may be an option.

onmete:
@tisnik @manstis I'm not entirely sure if it is a bug. Seems more like an intention. Anyway, I created a [PR ](https://github.com/meta-llama/llama-stack/pull/2575#issuecomment-3024434159)in llama-stack, lets see how that evolves.

onmete:
https://github.com/meta-llama/llama-stack/pull/2575 was merged so this PR can be closed.
Note ols supports more mime types than just `application/yaml`.

PR #160, opened 2025-07-01
LCORE-313: Fix image building with Python 3.12

matysek:
## Description

LCORE-313: Fix image building with Python 3.12

## Type of change

- [x] Bug fix


tisnik (approved):
LGTM!

Issue #123, opened 2025-06-24
[RFE] Move system_prompt to configuration


manstis:
**Is your feature request related to a problem? Please describe.**

The `system_prompt` is currently part of the `QueryRequest` model.

System Prompts can be very large and serialising them in every request is expensive.

It might be nice to add support for providing a `system_prompt` in the `ServiceConfiguration` object.

Issue #122, opened 2025-06-24
[RFE] Instantiating LlamaStackClient on each request is expensive


manstis:
**Is your feature request related to a problem? Please describe.**

`lightspeed-stack` instantiates a `LlamaStackClient` on each and every incoming request.

Construction of this, when used with `use_as_library_client: true` is a very expensive operation.

It might be better to instantiate the client _once_ and keep as a singleton for all operations.

**Additional context**
We're using a stack configuration that loads external providers.

The runtime stack implementation is recreated for every request and causes responses to be unnecessarily slow.

An alternative is to use an _external_ `llama-stack` service/server however that would require _us_ to maintain another Docker Container Image referencing the `llama-stack` libraries directly.

We have previously ran our own `llama-stack` instance however it led to (political) problems relating to _ownership_ of the `llama-stack` libraries when used in a _downstream_/product deployment.

manstis:
Fixed with https://github.com/lightspeed-core/lightspeed-stack/pull/187

PR #163, opened 2025-07-01
Do not create new session if conversation_id is provided

rawagner:
## Description

Before this change, a new session is created for every query, which makes the assistant hardly use-able.

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/src/app/endpoints/query.py
+++ b/src/app/endpoints/query.py
@@ -6,8 +6,10 @@ import logging
 import os
 from pathlib import Path
 from typing import Any
-from llama_stack_client.lib.agents.agent import Agent
 
+from cachetools import TTLCache  # type: ignore
+
+from llama_stack_client.lib.agents.agent import Agent
 from llama_stack_client import APIConnectionError
 from llama_stack_client import LlamaStackClient  # type: ignore
 from llama_stack_client.types import UserMessage  # type: ignore
@@ -32,6 +34,8 @@ from utils.suid import get_suid
 logger = logging.getLogger("app.endpoints.handlers")
 router = APIRouter(tags=["query"])
 
+# Global agent registry to persist agents across requests
+_agent_cache: TTLCache[str, Agent] = TTLCache(maxsize=1000, ttl=3600)
 
 query_response: dict[int | str, dict[str, Any]] = {
     200: {
@@ -56,16 +60,33 @@ def is_transcripts_enabled() -> bool:
     return not configuration.user_data_collection_configuration.transcripts_disabled
 
 
-def retrieve_conversation_id(query_request: QueryRequest) -> str:
-    """Retrieve conversation ID based on existing ID or on newly generated one."""
-    conversation_id = query_request.conversation_id
+def get_agent(
+    client: LlamaStackClient,
+    model_id: str,
+    system_prompt: str,
+    available_shields: list[str],
+    conversation_id: str | None,
+) -> tuple[Agent, str]:
+    """Get existing agent or create a new one with session persistence."""
+    if conversation_id is not None:
+        agent = _agent_cache.get(conversation_id)
+        if agent:
+            logger.debug("Reusing existing agent with key: %s", conversation_id)
+            return agent, conversation_id
 
-    # Generate a new conversation ID if not provided
-    if not conversation_id:
-        conversation_id = get_suid()
-        logger.info("Generated new conversation ID: %s", conversation_id)
-
-    return conversation_id
+    logger.debug("Creating new agent")
+    # TODO(lucasagomes): move to ReActAgent
+    agent = Agent(
+        client,
+        model=model_id,
+        instructions=system_prompt,
+        input_shields=available_shields if available_shields else [],
+        tools=[mcp.name for mcp in configuration.mcp_servers],
+        enable_session_persistence=True,
+    )
+    conversation_id = agent.create_session(get_suid())
+    _agent_cache[conversation_id] = agent
+    return agent, conversation_id
 
 
 @router.post("/query", responses=query_response)
@@ -83,8 +104,9 @@ def query_endpoint_handler(
         # try to get Llama Stack client
         client = get_llama_stack_client(llama_stack_config)
         model_id = select_model_id(client.models.list(), query_request)
-        conversation_id = retrieve_conversation_id(query_request)
-        response = retrieve_response(client, model_id, query_request, auth)
+        response, conversation_id = retrieve_response(
+            client, model_id, query_request, auth
+        )
 
         if not is_transcripts_enabled():
             logger.debug("Transcript collection is disabled in the configuration")
@@ -163,7 +185,7 @@ def retrieve_response(
     model_id: str,
     query_request: QueryRequest,
     token: str,
-) -> str:
+) -> tuple[str, str]:
     """Retrieve response from LLMs and agents."""
     available_shields = [shield.identifier for shield in client.shields.list()]
     if not available_shields:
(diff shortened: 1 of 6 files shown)


src/app/endpoints/query.py
```diff
@@ -181,8 +183,38 @@ def retrieve_response(
...
+            sessions = getattr(agent, 'sessions', [])
+            existing_session_id = None
+            for session_id in sessions:
+                if conversation_id in session_id:
+                    existing_session_id = session_id
+                    logger.debug("Found existing session: %s", existing_session_id)
+                    break
+
+            if existing_session_id:
+                session_id = existing_session_id
```
manstis, line 203: This is obsolete really.

`existing_session_id` is only set [here](https://github.com/lightspeed-core/lightspeed-stack/pull/163/files#diff-7f944302f13bc6873126bf121368f936473f714b87dec862b0d1a53e979e8c6dR198) and is set to `session_id`.

src/app/endpoints/query.py
```diff
@@ -181,8 +183,38 @@ def retrieve_response(
             ),
         },
     )
-    session_id = agent.create_session("chat_session")
-    logger.debug("Session ID: %s", session_id)
+
+    # Use conversation_id to determine session
+    conversation_id = query_request.conversation_id
```
manstis, line 188: A `conversation_id` is created [here](https://github.com/lightspeed-core/lightspeed-stack/blob/cccba036070de1f610c2ca668dd8acb72e8a949c/src/app/endpoints/query.py#L69) and returned from the `/query` endpoint [here](https://github.com/lightspeed-core/lightspeed-stack/blob/cccba036070de1f610c2ca668dd8acb72e8a949c/src/app/endpoints/query.py#L87).

src/app/endpoints/query.py
```diff
@@ -181,8 +183,38 @@ def retrieve_response(
...
+                logger.debug("Created new session: %s", session_id)
+
+        except Exception as e:
+            # If anything fails, fall back to creating a new session
+            logger.warning("Error checking existing sessions: %s", e)
+            session_id = agent.create_session(session_name)
+            logger.debug("Created fallback session: %s", session_id)
+    else:
+        # No conversation_id provided - create a new session
+        session_id = agent.create_session("chat_session")
```
manstis, line 215: You're effectively creating a new `conversation_id` here; but the `lightspeed-stack` code has already created one.

Therefore the `conversation_id` returned from `/query` will never match a `session_id` created here and [this code](https://github.com/lightspeed-core/lightspeed-stack/pull/163/files#diff-7f944302f13bc6873126bf121368f936473f714b87dec862b0d1a53e979e8c6dR205) will run on each request; creating more and more unmatched sessions.

manstis (commented):
I had a quick look out of curiosity really.

I think how a `conversation_id` is created needs some thought as it's really the `llama-stack` `session_id` now.

eranco74:
@onmete how is this done in OLS?

tisnik:
> @onmete how is this done in OLS?

there is no Llama Stack involved in OLS.
So just conversation_id is needed:
- when not provided in query endpoint, it is created
- when provided, it is used as is

Nothing super special there

rawagner:
Thanks for the quick review.

I've looked into this a bit further and realized that new Agent is created on every request - which results in having empty sessions. I've added a code to cache the Agents so that we can reuse the instances in following requests.

I dont think that is the solution you'd want to go with and I'm not familiar with the code-base - though at least it helps to demonstrate the problem ?

manstis:
@rawagner I think this is a step in the right direction.

Personally, if a `conversation_id` was not present in the `QueryRequest` I'd first create an `Agent` and `session`; returning the `session_id` as the `conversation_id`... they are one and the same thing just with different terms (`conversation_id`, IIRC, is a `road-core` term carried into `lightspeed-core` and `session_id` is a `llama-stack` term.. but they both mean the same thing).

You are correct that we need to re-use `Agent` for a _conversation_.

Looking at the `llama-stack` [source code](https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/inline/agents/meta_reference/agents.py#L91-L107) for an `Agent` the persistent store (for the _conversation_) is created when the `Agent` is instantiated. If we create a new `Agent` for each `QueryRequest` a new [`agent_id`](https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/inline/agents/meta_reference/agents.py#L95) is assigned and a [new persistent store](https://github.com/meta-llama/llama-stack/blob/main/llama_stack/providers/inline/agents/meta_reference/agents.py#L104-L107) created.... meaning we effectively loose any conversational history.

Similar changes will be needed for https://github.com/lightspeed-core/lightspeed-stack/blob/main/src/app/endpoints/streaming_query.py

I suspect this PR also addresses https://github.com/lightspeed-core/lightspeed-stack/issues/122

Finally, and conscious of my participation on `lightspeed-core`, I'm just a contributor like yourself. I am not part of the `lightspeed-core` team and therefore review/approval/rejection of this PR is really for the likes of @tisnik to decide.

I do however believe that this PR is important, addressing a serious flaw in the existing code.

tisnik:
@rawagner it looks nice now! You'd need rebase/resolve conflict - the same code was changed meantime. But it should be easy. Thanks a lot in advance!

manstis:
> it looks nice now! You'd need rebase/resolve conflict - the same code was changed meantime. But it should be easy. Thanks a lot in advance!

@rawagner @tisnik 

And we need the same changes made to the `streaming_query.py` endpoint handler :+1:

rawagner:
Looking into it. Thanks for the reviews!

rawagner:
For now, i've just quick-fixed the streaming_query. I am looking for a proper solution.

src/app/endpoints/query.py
```diff
@@ -48,16 +50,33 @@ def is_transcripts_enabled() -> bool:
...
+        if agent:
+            logger.debug("Reusing existing agent with key: %s", conversation_id)
+            return agent, conversation_id
+
+    logger.debug("Creating new agent")
+    # TODO(lucasagomes): move to ReActAgent
+    agent = Agent(
+        client,
+        model=model_id,
+        instructions=system_prompt,
```
manstis, line 72: This represents an interesting scenario I missed earlier.

`system_prompt` is part of the `QueryRequest` class that is used _per request_.

The `Agent` is now cached with whatever `system_prompt` was supplied on the *first* request.

Meaning `QueryRequest.system_prompt` effectively becomes redundant.

I think we should probably remove `instructions=system_prompt` from here and ... <see my other comment>

manstis (commented):
The changes as they are are _generally_ fine :+1: 

However it does introduce an edge-case that could be problematic. I propose a solution.

The edge-case would only manifest itself if _different_ requests had _different_ `system_prompt`. 

Whether:

- It's considered an unlikely scenario to worry about
- It should be fixed in this PR
- It should be fixed in a new PR

I leave to you.

src/app/endpoints/streaming_query.py
```diff
@@ -31,6 +33,37 @@
...
+        agent = _agent_cache.get(conversation_id)
+        if agent:
+            logger.debug("Reusing existing agent with key: %s", conversation_id)
+            return agent, conversation_id
+
+    logger.debug("Creating new agent")
+    agent = AsyncAgent(
+        client,  # type: ignore[arg-type]
+        model=model_id,
+        instructions=system_prompt,
```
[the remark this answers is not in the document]
eranco74: Why remove this?
Why should we add the system prompt to each message instead of the Agent instructions?
https://github.com/meta-llama/llama-stack/blob/4bca4af3e42ea2290973e97555fb8736246c62d3/llama_stack/providers/inline/agents/meta_reference/agent_instance.py#L170

src/app/endpoints/query.py
eranco74: TBH, it seems strange that the system prompt is passed in the request.
* Also relates to https://github.com/lightspeed-core/lightspeed-stack/issues/123

I think that we should keep this and in case the `query_request.system_prompt` isn't empty we shold add it to the mesage (as you suggested.

Unsure whether that is the right thing to do here, perhaps it's better to differ this and solve it in another PR (along with #123.

tisnik (changes requested):
Please:

1) rebase
2) remove pdm.lock completely, we switched to `uv` to meantime (sorry, too many changes occurred in the last week)

PR #177, opened 2025-07-03
Add media_type support for streaming_query

manstis:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement

None.

## Testing

### DIFF
--- a/src/app/endpoints/streaming_query.py
+++ b/src/app/endpoints/streaming_query.py
@@ -134,6 +134,7 @@ async def streaming_query_endpoint_handler(
     model_id = select_model_id(await client.models.list(), query_request)
     conversation_id = retrieve_conversation_id(query_request)
     response = await retrieve_response(client, model_id, query_request)
+    media_type = query_request.media_type
 
     async def response_generator(turn_response: Any) -> AsyncIterator[str]:
         """Generate SSE formatted streaming response."""
@@ -141,7 +142,8 @@ async def streaming_query_endpoint_handler(
         complete_response = ""
 
         # Send start event
-        yield stream_start_event(conversation_id)
+        if media_type == constants.MEDIA_TYPE_JSON:
+            yield stream_start_event(conversation_id)
 
         async for chunk in turn_response:
             if event := stream_build_event(chunk, chunk_id):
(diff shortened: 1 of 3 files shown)


src/app/endpoints/streaming_query.py
```diff
@@ -134,14 +134,16 @@ async def streaming_query_endpoint_handler(
...
 
     async def response_generator(turn_response: Any) -> AsyncIterator[str]:
         """Generate SSE formatted streaming response."""
         chunk_id = 0
         complete_response = ""
 
         # Send start event
-        yield stream_start_event(conversation_id)
+        if media_type == constants.MEDIA_TYPE_JSON:
+            yield stream_start_event(conversation_id)
```
umago, line 146: Should we do the same around the stream_end_event() on L156 ?

src/models/requests.py
```diff
@@ -80,6 +80,7 @@ class QueryRequest(BaseModel):
     model: Optional[str] = None
     system_prompt: Optional[str] = None
     attachments: Optional[list[Attachment]] = None
+    media_type: Optional[str] = constants.MEDIA_TYPE_JSON
```
umago, line 83: Not sure we should be adding media_type as part of this class or create another one for the streaming query that inherits from this and add that field. 

Saying this because I don't think we will be support media_type as a parameter for /query.

So I would suggest something like:

```
class StreamingQueryRequest(QueryRequest):
    media_type = Optional[str] = constants.MEDIA_TYPE_JSON

    @model_validator(mode="after")
    def validate_media_type(self) -> Self:
        # do the validation of the media_type field here
```

src/models/requests.py
```diff
@@ -80,6 +80,7 @@ class QueryRequest(BaseModel):
...
 
     # provides examples for /docs endpoint
     model_config = {
```
umago, line 86: We also need to add the new media_type field to this "model_config" attribute as well as creating validation for it:

``` 
    @model_validator(mode="after")
    def validate_media_type(self) -> Self:
        ....
```

umago (changes requested):
Thanks @manstis! Comments inline

manstis:
Closing.. Ansible Lightspeed only uses `application/json` and therefore the changes herein are not needed.

PR #188, opened 2025-07-04
LCORE-311: Update readme with image tag info

matysek:
## Description

LCORE-311: Update readme with image tag info

## Type of change

- [x] Documentation Update


tisnik (approved):
nice one!

PR #189, opened 2025-07-04
LCORE-283: Introduce authentication to LS-stack API

umago:
## Description

This patch introduce authentication mechanisms to the Lightspeed-stack API.

Following what is in road-core/service, all 3 methods have been reintroduced:
   * noop - Which is just a No-op method, used for development
   * noop-with-token - Similar to No-op but require a Authorization Header to be passed as part of the request. For development use only
   * k8s - Which authenticates with Kubernetes/OCP

A few notes about this patch:

1. noop is the default authentication method. In road-core/service the default method was "k8s" but, since this is not a "k8s" first project
2. The "skip_user_id_check" was removed from the AuthInterface return values because it wasn't used anywhere
3. The "k8s.py" file contains a lot of pyright warnings, 58 to be exact. This file was basically c&p from the existing one in road-core/service.So I left those to be fixed later. We may want to refactor how we do things there.
4. No dev_config has been introduced as part of this patch

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement

- Related Issue #LCORE-283
- Closes #

## Testing
- Unit-tests with extensive coverage
- Manually tested


umago:
> it looks ok-ish. Would be nice to be "clean" on CI, but probably ok for now (I can fix it later I guess)
> 
> @manstis PTAL - it will affect you too

@tisnik if you want I can work on that, what needs to be cleaned up ?

tisnik:
> > it looks ok-ish. Would be nice to be "clean" on CI, but probably ok for now (I can fix it later I guess)
> > @manstis PTAL - it will affect you too
> 
> @tisnik if you want I can work on that, what needs to be cleaned up ?

When I looked at CI, there were some linter failures. Now are gone -> we are ok. thanks

umago:
@tisnik ah sorry, it was conflict from other patches that merged before. I fixed it now

tisnik (approved):
nice one

PR #202, opened 2025-07-08
LCORE-312: Support MCP Headers

ldjebran:
## Description

issue: https://issues.redhat.com/browse/LCORE-312
this supports the client providing the headers to supply to mcp servers via llama-stack in the form of a dictionary like:
{
    "mcp-server-url-1": {"HEADER-1": "HEADER-1-VALUE-1", "HEADER-1": "HEADER-1-VALUE-2"},
    "mcp-server-url-2": {"HEADER-2": "HEADER-2-VALUE"}
}

Suppose each different mcp server support it's own authentication type, or needs some headers to be supplied, this PR will allow to fine tune which headers may be passed to which mcp server.

For the moment did not moved the extra_headers to create_turn function as the AsyncAgent is not supporting that. 

moved the toolgroups to create_turn function as when declared at the agent, the headers are not sent to mcp servers the first time llama-stack is listing the tools from mcp servers, that may lead to authentication failure.

The optimal is to move also the headers to be declared in create_turn function and not at the agent level, but this should be done in it's own task ticket. as we weed to implement the same for AsyncAgent. This will allow us to have one agent instance for all sessions and remove the agent caching. 

## Type of change

- [ ] Refactor
- [X] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [x] Unit tests improvement

- Closes # https://issues.redhat.com/browse/LCORE-312

- [X] I have performed a self-review of my code.
- [X] If it is a core feature, I have added thorough tests.

## Testing
Tested with 2 mcp servers that support authentication, the mcp servers does not allow tools listing without authentication too, put in configuration file lightspeed-stack.yaml
```
...
mcp_servers:
- name: mcp::aap-controller
  provider_id: model-context-protocol
  url: http://localhost:8004/sse
- name: mcp::aap-gateway
  provider_id: model-context-protocol
  url: http://localhost:8003/sse
...
```

The llama-stack was running as an external service by default using model "granite-3.3-8b-instruct" via vllm container runing in openshift 

Created 2 scripts one for streaming and one for non streaming endpoints:

https://github.com/user-attachments/assets/623bf79c-9d90-44ca-9938-2ec857baa9f6

### DIFF
--- a/src/app/endpoints/streaming_query.py
+++ b/src/app/endpoints/streaming_query.py
@@ -22,6 +22,7 @@ from configuration import configuration
 from models.requests import QueryRequest
 from utils.endpoints import check_configuration_loaded, get_system_prompt
 from utils.common import retrieve_user_id
+from utils.mcp_headers import mcp_headers_dependency
 from utils.suid import get_suid
 
 
@@ -61,7 +62,6 @@ async def get_agent(
         model=model_id,
         instructions=system_prompt,
         input_shields=available_shields if available_shields else [],
-        tools=[mcp.name for mcp in configuration.mcp_servers],
         enable_session_persistence=True,
     )
     conversation_id = await agent.create_session(get_suid())
@@ -187,6 +187,7 @@ async def streaming_query_endpoint_handler(
     _request: Request,
     query_request: QueryRequest,
     auth: Any = Depends(auth_dependency),
+    mcp_headers: dict[str, dict[str, str]] = Depends(mcp_headers_dependency),
 ) -> StreamingResponse:
     """Handle request to the /streaming_query endpoint."""
     check_configuration_loaded(configuration)
@@ -199,7 +200,11 @@ async def streaming_query_endpoint_handler(
         client = await get_async_llama_stack_client(llama_stack_config)
         model_id = select_model_id(await client.models.list(), query_request)
         response, conversation_id = await retrieve_response(
-            client, model_id, query_request, auth
+            client,
+            model_id,
+            query_request,
+            auth,
+            mcp_headers=mcp_headers,
         )
         metadata_map: dict[str, dict[str, Any]] = {}
 
@@ -255,6 +260,7 @@ async def retrieve_response(
     model_id: str,
     query_request: QueryRequest,
     token: str,
+    mcp_headers: dict[str, dict[str, str]] | None = None,
 ) -> tuple[Any, str]:
     """Retrieve response from LLMs and agents."""
     available_shields = [shield.identifier for shield in await client.shields.list()]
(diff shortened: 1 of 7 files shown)


src/utils/mcp_headers.py
```diff
@@ -0,0 +1,44 @@
...
+    Returns:
+        The mcp headers dictionary, or empty dictionary if not found or on json decoding error
+    """
+    mcp_headers_string = request.headers.get("MCP-HEADERS", "")
+    mcp_headers = {}
+    if mcp_headers_string:
+        try:
+            mcp_headers = json.loads(mcp_headers_string)
+        except json.decoder.JSONDecodeError as e:
+            logger.debug("MCP headers decode error: %s", e)
```
umago, line 39: Log should use error level

src/utils/mcp_headers.py
```diff
@@ -0,0 +1,44 @@
...
+
+        if not isinstance(mcp_headers, dict):
+            logger.debug("MCP headers wrong format: %s", mcp_headers)
```
umago, line 42: Ditto, error level

src/utils/mcp_headers.py
```diff
@@ -0,0 +1,44 @@
...
+def extract_mcp_headers(request: Request) -> dict[str, dict[str, str]]:
+    """Extract mcp headers from MCP-HEADERS header.
+
+    Args:
+        request: The FastAPI request object
+
+    Returns:
+        The mcp headers dictionary, or empty dictionary if not found or on json decoding error
+    """
+    mcp_headers_string = request.headers.get("MCP-HEADERS", "")
```
umago, line 33: Just so I understand, we are passing these MCP headers via this "MCP-HEADERS" as part of the request but, don't we want to have it in the [MCP configuration class](https://github.com/lightspeed-core/lightspeed-stack/blob/2bdda1afc7ca3534b6fbeef05804a896c759ec4c/src/models/config.py#L49-L54) ?

umago (commented):
Thanks for the PR, left a question inline

src/app/endpoints/streaming_query.py
```diff
@@ -285,8 +291,10 @@ async def retrieve_response(
         query_request.conversation_id,
     )
 
-    mcp_headers = {}
-    if token:
+    # preserve compatibility when mcp_headers is not provided
+    if mcp_headers is None:
+        mcp_headers = {}
+    if not mcp_headers and token:
```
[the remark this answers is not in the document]
ldjebran: do not thing this needs changes , after I explained that we need to init the mcp_header kwarg

src/utils/mcp_headers.py
ldjebran: really do not to log this error, as not that much important, I tend to be silent on prod , on this kind of errors, but if you think this is a blocker will update it. I have no preferences on this.

src/utils/mcp_headers.py
ldjebran: same as as previous response

src/utils/mcp_headers.py
ldjebran: updated, with error level

src/utils/mcp_headers.py
ldjebran: updated with error level but showing only the errored type.

umago (approved):
Thanks! LGTM

tisnik (approved):
it looks good, thank you

PR #205, opened 2025-07-08
Fix llama_stack_client log levels

Akrog:
## Description

Log levels on the llama_stack_client are currently broken and setting
`LLAMA_STACK_LOG=debug` doesn't work as it should.  Log levels are
automatically changed to WARNING.

This is caused by us blindly loading the `llama_stack` library, which
initializes logging to WARNING except a subset of modules that can be
modified using the `LLAMA_STACK_LOGGING` env var.

With this patch we only import `llama_stack` when we are actually going
to use it, so that when we use the normal llama-stack client we can set
its logging level.

## Type of change

- [x] Bug fix

## Testing

Just run lightspeed-stack with a remote llama-stack and set the `LLAMA_STACK_LOGGING` to `debug` and see that you don't get debug logs, but you do once you apply this patch.

## Summary by CodeRabbit

* **Refactor**
  * Improved how certain components are loaded internally to optimize resource usage. No changes to visible features or functionality.

### DIFF
--- a/src/client.py
+++ b/src/client.py
@@ -4,10 +4,6 @@ import logging
 
 from typing import Optional
 
-from llama_stack.distribution.library_client import (
-    AsyncLlamaStackAsLibraryClient,  # type: ignore
-    LlamaStackAsLibraryClient,  # type: ignore
-)
 from llama_stack_client import AsyncLlamaStackClient, LlamaStackClient  # type: ignore
 from models.config import LLamaStackConfiguration
 from utils.types import Singleton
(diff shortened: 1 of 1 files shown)


tisnik (changes requested):
it looks ok, thank you. Just please rebase your PR

umago (commented):
Code-wise it looks fine, just need to fix the current issues with CI

umago:
@Akrog r u going to take a look at this ? I think we can merge it once the linters issues are fixed.

tisnik:
@Akrog Hello, is this one still relevant?

tisnik:
@Akrog  should we continue improving this PR or should it be closed?

PR #216, opened 2025-07-09
WIP: Add an option to disable query system prompt

omertuc:
Since LCORE-293 (e60d7df1beca37a5909e5b066e39162c8fab889b), users can
set specify the system prompt in the config, but users can still
override it with their queries. This may not be desired in some cases,
so this commit adds an option to disable the query system prompt and
always use the configured system prompt (or the default one if none is
set).

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement

- Related Issue #

LCORE-293

- Closes #

## Testing


src/models/config.py
```diff
@@ -128,11 +128,12 @@ def check_authentication_model(self) -> Self:
 class Customization(BaseModel):
     """Service customization."""
 
+    disable_query_system_prompt: Optional[bool] = False
```
manstis, line 131: Nit picking.

Most of other `bool` fields are named `xxxx_enabled` or `xxxx_disabled`.

For consistency it might be nicer to name this `query_system_prompt_disabled`.

- https://github.com/lightspeed-core/lightspeed-stack/blob/bc717112894198f032f3011a7061f02e7ed52c0b/src/models/config.py#L31
- https://github.com/lightspeed-core/lightspeed-stack/blob/bc717112894198f032f3011a7061f02e7ed52c0b/src/models/config.py#L91
- https://github.com/lightspeed-core/lightspeed-stack/blob/bc717112894198f032f3011a7061f02e7ed52c0b/src/models/config.py#L93

src/utils/endpoints.py
```diff
@@ -19,8 +19,12 @@ def check_configuration_loaded(configuration: AppConfig) -> None:
 def get_system_prompt(query_request: QueryRequest, configuration: AppConfig) -> str:
     """Get the system prompt: the provided one, configured one, or default one."""
     # system prompt defined in query request has precendence
-    if query_request.system_prompt:
-        return query_request.system_prompt
+    if (
+        configuration.customization is not None
```
tisnik, line 23: it won't be `None`, you meant `False` perhaps?

tisnik (changes requested):

manstis (approved):
LGTM :+1: 

Just a nit-pick for you to consider.

umago:
@eranco74 when I was doing the /query, I mimic what was available in the road-core/service /query endpoint. And the system_prompt was a parameter.

Now, if nobody uses it and if we see it as a problem. Let's get rid of it before there's any release instead of having 3 different ways of configurating the system_prompt ? 

What you folks think ? I think we should make things simple and straight-forward, adding a new config to disable something in the API makes the use of that strange.

With this patch, when the config is enabled and the user make a request to the /query endpoint with the system_prompt set, the service will return 200 saying the request was accepted but in reality part of it was ignored. It sounds off to me.

src/utils/endpoints.py
```diff
@@ -19,8 +19,12 @@ def check_configuration_loaded(configuration: AppConfig) -> None:
 def get_system_prompt(query_request: QueryRequest, configuration: AppConfig) -> str:
```
umago, line 19: So now, if user makes a request to the API with the system_prompt set. The service will return 200 saying it was accepted but, if this new option is enabled. The request was just partially accept right ? As part of it will be ignored.

I don't like it, I think that we should just get rid of the system_prompt parameter in the /query endpoint. Make the service simple, because now we have too many ways of configuring/overriding the same thing.

umago:
Ok so speaking with other folks, the "system_prompt" parameter does have some usage. We then agreed on raising an error when the option that disable the system_prompt is enabled and a user makes a request with the system_prompt set.

That would solve the issue and makes the API usage clear from the user perspective. Thanks for the patch!

manstis:
FWIW, the Ansible Chatbot makes use of `road-core`'s `QueryRequest.system_prompt` to override the `system_prompt` without having to build a new backend service image with it statically configured.. It has been very useful when working on "prompt engineering" to be able to quickly try new `system_prompt`'s during _development_.

There are probably other approaches that could be taken if it was removed from `lightspeed-stack`.

umago:
Thanks @manstis for the feedback. Yeah, I can see the benefit of it now. Speaking with @omertuc, we decided to throw an error when the option to disable query_system_prompt is set and the user sends a request with a system_prompt.

I think that's fair because, we don't just sneakely change or ignore parts of the user request and also we keep the endpoint for the cases like yours. Does that work for you ?

src/utils/endpoints.py
omertuc: I meant None. I just copied from the existing code. Was the existing code wrong?

src/utils/endpoints.py
```diff
@@ -18,12 +18,24 @@ def check_configuration_loaded(configuration: AppConfig) -> None:
...
     """Get the system prompt: the provided one, configured one, or default one."""
-    # system prompt defined in query request has precendence
+    system_prompt_disabled = (
+        configuration.customization is not None
+        and configuration.customization.disable_query_system_prompt
+    )
+    if system_prompt_disabled and query_request.system_prompt:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
```
umago, line 27: Lovely, 422 makes sense!

src/utils/endpoints.py
umago: Seems to match with L40 of this same file

umago (commented):
Other than the pylint error, the code LGTM now

PR #244, opened 2025-07-15
LCORE-354 Implement the /conversations endpoint

eranco74:
Add /conversations endpoint for conversation history management

## Description

- Add GET /v1/conversations/{conversation_id} to retrieve conversation history
- Add DELETE /v1/conversations/{conversation_id} to delete conversations
- Use llama-stack client.agents.session.retrieve and .delete methods
- Map conversation ID to agent ID for LlamaStack operations
- Add ConversationResponse and ConversationDeleteResponse models
- Include conversations router in main app routing
- Maintain consistent error handling and authentication patterns

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement

LCORE-354

- [] If it is a core feature, I have added thorough tests.

## Testing

## Summary by CodeRabbit

* **New Features**
  * Introduced API endpoints to retrieve and delete conversation history with comprehensive responses and error handling.
  * Added response formats for conversation retrieval and deletion actions to enhance API documentation.
  * Registered conversation-related routes under the versioned API prefix.
* **Bug Fixes**
  * Enhanced mapping between conversation IDs and agent IDs for improved session management reliability.
* **Tests**
  * Updated tests to cover the new conversations API endpoints and verify their integration within the application.

### DIFF
--- a/src/app/endpoints/conversations.py
+++ b/src/app/endpoints/conversations.py
@@ -1,0 +1,263 @@
+"""Handler for REST API calls to manage conversation history."""
+
+import logging
+from typing import Any
+
+from llama_stack_client import APIConnectionError, NotFoundError
+
+from fastapi import APIRouter, HTTPException, status, Depends
+
+from client import LlamaStackClientHolder
+from configuration import configuration
+from models.responses import ConversationResponse, ConversationDeleteResponse
+from auth import get_auth_dependency
+from utils.endpoints import check_configuration_loaded
+from utils.suid import check_suid
+
+logger = logging.getLogger("app.endpoints.handlers")
+router = APIRouter(tags=["conversations"])
+auth_dependency = get_auth_dependency()
+
+conversation_id_to_agent_id: dict[str, str] = {}
+
+conversation_responses: dict[int | str, dict[str, Any]] = {
+    200: {
+        "conversation_id": "123e4567-e89b-12d3-a456-426614174000",
+        "session_data": {
+            "session_id": "123e4567-e89b-12d3-a456-426614174000",
+            "turns": [],
+            "started_at": "2024-01-01T00:00:00Z",
+        },
+    },
+    404: {
+        "detail": {
+            "response": "Conversation not found",
+            "cause": "The specified conversation ID does not exist.",
+        }
+    },
+    503: {
+        "detail": {
+            "response": "Unable to connect to Llama Stack",
+            "cause": "Connection error.",
+        }
+    },
+}
+
+conversation_delete_responses: dict[int | str, dict[str, Any]] = {
+    200: {
+        "conversation_id": "123e4567-e89b-12d3-a456-426614174000",
+        "success": True,
+        "message": "Conversation deleted successfully",
+    },
+    404: {
+        "detail": {
+            "response": "Conversation not found",
+            "cause": "The specified conversation ID does not exist.",
+        }
+    },
+    503: {
+        "detail": {
+            "response": "Unable to connect to Llama Stack",
+            "cause": "Connection error.",
+        }
+    },
+}
+
+
+def simplify_session_data(session_data: Any) -> list[dict[str, Any]]:
+    """Simplify session data to include only essential conversation information.
+
+    Args:
+        session_data: The full session data from llama-stack
+
+    Returns:
+        Simplified session data with only input_messages and output_message per turn
+    """
+    session_dict = session_data.model_dump()
+    # Create simplified structure
+    chat_history = []
+
+    # Extract only essential data from each turn
+    for turn in session_dict.get("turns", []):
+        # Clean up input messages
+        cleaned_messages = []
+        for msg in turn.get("input_messages", []):
+            cleaned_msg = {
+                "content": msg.get("content"),
+                "type": msg.get("role"),  # Rename role to type
+            }
+            cleaned_messages.append(cleaned_msg)
+
+        # Clean up output message
+        output_msg = turn.get("output_message", {})
+        cleaned_messages.append(
+            {
+                "content": output_msg.get("content"),
+                "type": output_msg.get("role"),  # Rename role to type
+            }
+        )
+
+        simplified_turn = {
+            "messages": cleaned_messages,
+            "started_at": turn.get("started_at"),
+            "completed_at": turn.get("completed_at"),
+        }
+        chat_history.append(simplified_turn)
+
+    return chat_history
+
+
+@router.get("/conversations/{conversation_id}", responses=conversation_responses)
+def get_conversation_endpoint_handler(
+    conversation_id: str,
+    _auth: Any = Depends(auth_dependency),
+) -> ConversationResponse:
+    """Handle request to retrieve a conversation by ID."""
+    check_configuration_loaded(configuration)
+
+    # Validate conversation ID format
+    if not check_suid(conversation_id):
+        logger.error("Invalid conversation ID format: %s", conversation_id)
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail={
+                "response": "Invalid conversation ID format",
+                "cause": f"Conversation ID {conversation_id} is not a valid UUID",
+            },
+        )
+
+    agent_id = conversation_id_to_agent_id.get(conversation_id)
+    if not agent_id:
+        logger.error("Agent ID not found for conversation %s", conversation_id)
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail={
+                "response": "conversation ID not found",
+                "cause": f"conversation ID {conversation_id} not found!",
+            },
+        )
+
+    logger.info("Retrieving conversation %s", conversation_id)
+
+    try:
+        client = LlamaStackClientHolder().get_client()
+
+        session_data = client.agents.session.retrieve(
+            agent_id=agent_id, session_id=conversation_id
+        )
+
+        logger.info("Successfully retrieved conversation %s", conversation_id)
+
+        # Simplify the session data to include only essential conversation information
+        chat_history = simplify_session_data(session_data)
+
+        return ConversationResponse(
+            conversation_id=conversation_id,
+            chat_history=chat_history,
+        )
+
+    except APIConnectionError as e:
+        logger.error("Unable to connect to Llama Stack: %s", e)
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+            detail={
+                "response": "Unable to connect to Llama Stack",
+                "cause": str(e),
+            },
+        ) from e
+    except NotFoundError as e:
+        logger.error("Conversation not found: %s", e)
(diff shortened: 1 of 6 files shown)


eranco74:
@coderabbitai

eranco74:
[coderabbitai](https://github.com/apps/coderabbitai) review

src/models/responses.py
```diff
@@ -298,3 +298,75 @@ class ForbiddenResponse(UnauthorizedResponse):
...
+                    "conversation_id": "123e4567-e89b-12d3-a456-426614174000",
+                    "session_data": {
+                        "session_id": "123e4567-e89b-12d3-a456-426614174000",
+                        "turns": [],
+                        "started_at": "2024-01-01T00:00:00Z",
+                    },
+                }
+            ]
+        }
+    }
```
umago, line 336: In order to keep the API compatibility with road-core, we need to return the same structure from `GET .../conversations/<id>`.

This return model doesn't seem compatible with road-core [0][1]

[0] https://github.com/road-core/service/blob/9d65d15a4d1dec47e5aac15ee86fef39db975006/ols/app/models/models.py#L220-L247 

[1] https://github.com/road-core/service/blob/9d65d15a4d1dec47e5aac15ee86fef39db975006/ols/app/endpoints/conversations.py#L62C6-L62C25

src/models/responses.py
```diff
@@ -298,3 +298,75 @@ class ForbiddenResponse(UnauthorizedResponse):
...
+        "json_schema_extra": {
+            "examples": [
+                {
+                    "conversation_id": "123e4567-e89b-12d3-a456-426614174000",
+                    "success": True,
+                    "message": "Conversation deleted successfully",
+                }
+            ]
+        }
+    }
```
umago, line 372: Ditto, re: keeping it compatible with road-core [0].

For this one, I don't think adding the conversation_id and the success field is a problem. But I think we need to rename `message` to `response` in order to keep the compatibility.

[0] https://github.com/road-core/service/blob/9d65d15a4d1dec47e5aac15ee86fef39db975006/ols/app/models/models.py#L287-L310

src/app/endpoints/conversations.py
```diff
@@ -0,0 +1,203 @@
...
+            agent_id=agent_id, session_id=conversation_id
+        )
+
+        logger.info("Successfully retrieved conversation %s", conversation_id)
+
+        return ConversationResponse(
+            conversation_id=conversation_id,
+            session_data=(
+                session_data.model_dump()
+                if hasattr(session_data, "model_dump")
```
umago, line 112: Is there any case where the response won't be a Session [0] object ?

I think we should assume it's a Session [0] instance here.

[0] https://github.com/meta-llama/llama-stack-client-python/blob/b6c6c5ed7940bb625665d50f88ff7ea9d734e100/src/llama_stack_client/resources/agents/session.py#L234C10-L234C17

src/app/endpoints/conversations.py
```diff
@@ -0,0 +1,203 @@
...
+        logger.error("Unable to connect to Llama Stack: %s", e)
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+            detail={
+                "response": "Unable to connect to Llama Stack",
+                "cause": str(e),
+            },
+        ) from e
+    except Exception as e:
+        # Handle case where session doesn't exist or other errors
```
umago, line 127: Is there a way to handle the 404 separated from others ?

404 is a very specific error. For anything else, we should raise 500 instead.

src/app/endpoints/conversations.py
```diff
@@ -0,0 +1,203 @@
...
```
umago, line 195: ditto: re: separate 404 from 500 errors

umago (changes requested):
Thanks for the patch! Comments inline

src/models/responses.py
eranco74: Does it have to be compatible?
What llama-stack is returning seems a lot better than I see in the link above.
input_message and output_message seems better than content + type.
Not to mention the timestamps.
I agree that I should remove a few fields (steps, tool_execution, etc), but is it a requirement to align to what we have in road-core?

umago: Let's see what others say, from my understand, teams like Ansible are trying to replace the road-core/service with this new llama-stack implementation without having to change their client. @manstis please correct me if I am wrong.

If that's the case I think we will have to make it backward compatible.

That said, I do agree on cleaning up a bit the llama-stack output to remove the noise and I do like the timestamps too, they are useful.

I don't think the problem is adding new fields, we can totally do that as long as we keep the structure and old fields as before (assuming we want to keep it backward compatible).

manstis: @umago we don't use `/conversations` in `road-core` so are unaffected if there's backwards compatibility issues.

eranco74: Thanks @manstis.
In order to add the timestamps, I need to nest the content inside the messages list.
While we break compatibility :smile: Can we change type back to role?
Let me know who else we should consult on this?

@tisnik?

eranco74: Althouh I prefer the messages list since it provides better logical grouping of the messages. Better represent the conversation flow and makeing it easier to manipulate the output for analysis. 
The original API flattens this context.

src/app/endpoints/conversations.py
```diff
@@ -0,0 +1,263 @@
...
+from models.responses import ConversationResponse, ConversationDeleteResponse
+from auth import get_auth_dependency
+from utils.endpoints import check_configuration_loaded
+from utils.suid import check_suid
+
+logger = logging.getLogger("app.endpoints.handlers")
+router = APIRouter(tags=["conversations"])
+auth_dependency = get_auth_dependency()
+
+conversation_id_to_agent_id: dict[str, str] = {}
```
tisnik, line 21: just warning:

1) this will work only when the service will be scaled to one replica. Otherwise the two consecutive REST API calls can be served by different replicas and thus with different mapping

2) what about service restart? This info get lost

3) also it can grow a lot if we store this in memory (only)

ATM the lcore has no implementation of any permanent storage. Probably we can merge this one and create a task to allow lcore to store such information into SQLite/Redis/serialized JSON (?)

@manstis ?

manstis: @tisnik I share your concerns (and have the same concerns about the "[Agent Cache](https://github.com/lightspeed-core/lightspeed-stack/blob/main/src/app/endpoints/query.py#L41)"). IDK whether there's a JIRA to consider the scalability of LSC? It seems there are some features being added that may bite Lightspeed Teams later on.

@eranco74 It's very well documenting the gap; but IDK if adding it to a JIRA that is set to be closed is the best place?

Perhaps a follow-up LCORE JIRA would be a better place so it's visible on JIRA boards for things "to be done".

umago: I agree on creating a separated Jira ticket about the issue here

src/models/responses.py
umago: Thanks @manstis for clarifying. Our team also doesn't rely on the backward compatibility for this endpoint. Perhaps one last thing we can do is to ask in the internal slack if any of the teams relies on it. If no one cares, yeah, I'm all about changing it to something new and better

src/app/endpoints/conversations.py
tisnik: @eranco74 yes pls, create new ticket and if its for LCORE-2 (I guess so) and version 0.2.0 (very probably) we can refine it today PM

src/app/endpoints/conversations.py
umago: I don't think this has been resolved ?

src/app/endpoints/conversations.py
umago: ditto

src/app/endpoints/conversations.py
```diff
@@ -0,0 +1,263 @@
...
+            detail={
+                "response": "Conversation not found",
+                "cause": f"Conversation {conversation_id} could not be retrieved: {str(e)}",
+            },
+        ) from e
+    except Exception as e:
+        # Handle case where session doesn't exist or other errors
+        logger.error("Error retrieving conversation %s: %s", conversation_id, e)
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
```
umago, line 181: Should this be 500 ? 404 is being handled at L168

src/app/endpoints/conversations.py
```diff
@@ -0,0 +1,263 @@
...
+            detail={
+                "response": "Conversation not found",
+                "cause": f"Conversation {conversation_id} could not be deleted: {str(e)}",
+            },
+        ) from e
+    except Exception as e:
+        # Handle case where session doesn't exist or other errors
+        logger.error("Error deleting conversation %s: %s", conversation_id, e)
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
```
umago, line 258: Should this be 500 ? 404 is being handled at L248

src/app/endpoints/conversations.py
```diff
@@ -0,0 +1,263 @@
...
+            detail={
+                "response": "Conversation not found",
```
umago, line 260: Unknown error

src/app/endpoints/conversations.py
```diff
@@ -0,0 +1,263 @@
...
+                "cause": f"Conversation {conversation_id} could not be retrieved: {str(e)}",
+            },
+        ) from e
+    except Exception as e:
+        # Handle case where session doesn't exist or other errors
+        logger.error("Error retrieving conversation %s: %s", conversation_id, e)
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail={
+                "response": "Conversation not found",
```
umago, line 183: Unknown error

umago (changes requested):

eranco74:
[Akrog](https://github.com/Akrog) I agree it's not GA, but it's a start.

> This means that each turn will be completely separated and the internal conversation history in llama-stack will not provide previous context.

you have it wrong, we currently use agent_cache to reuse the same agent  for the same conversation.
You can see the multiple turns in output I pasted in the first comment (the output changed since but the multiple turns is there)
 
> Also, keeping the conversation_id_to_agent_id mapping in lightspeed-stack memory is insufficient, because a restart of the pod will lose the service's ability to get information.

That is currect, we will handle that in https://issues.redhat.com/browse/LCORE-373

>We are also not persisting the sessions in the llama-stack server, because we are not passing enable_session_persistence=True to the Agent on instantiation.

We are passing enable_session_persistence=True
see https://github.com/search?q=repo%3Alightspeed-core%2Flightspeed-stack%20enable_session_persistence&type=code

umago (approved):
Thanks LGTM now

PR #278, opened 2025-07-23
LCORE-175: Define central storage path config option for user data collection

onmete:
## Description

Create a single storage path. The feedback/transcripts/XYZ shouldn't be configurable, but hardcoded in lightspeed-core to ensure the archive's inner structure.

While it won't break the planned data ingestion, the current OLS reporting depends on this archive structure. So, unless there is a request to change it or have it configurable (not the scenario for a new datapoint inside the archive), let's remove it from configuration.

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement

https://issues.redhat.com/browse/LCORE-175

## Testing

## Summary by CodeRabbit

* **Documentation**
  * Updated documentation and configuration examples to use a unified user data directory for feedback and transcripts storage.

* **Refactor**
  * Simplified configuration by replacing separate feedback and transcripts storage paths with a single user data directory.
  * Updated internal logic and tests to use the new unified storage configuration.

* **Tests**
  * Revised and removed outdated tests to align with the new storage configuration approach.
  * Added tests to verify correct behavior with the unified user data directory.

### DIFF
--- a/src/app/endpoints/feedback.py
+++ b/src/app/endpoints/feedback.py
@@ -109,19 +109,18 @@ def store_feedback(user_id: str, feedback: dict) -> None:
         feedback: The feedback to store.
     """
     logger.debug("Storing feedback for user %s", user_id)
-    # Creates storage path only if it doesn't exist. The `exist_ok=True` prevents
-    # race conditions in case of multiple server instances trying to set up storage
-    # at the same location.
-    storage_path = Path(
-        configuration.user_data_collection_configuration.feedback_storage or ""
-    )
-    storage_path.mkdir(parents=True, exist_ok=True)
+
+    # Create feedback directory if it doesn't exist
+    feedback_dir = Path(
+        configuration.user_data_collection_configuration.feedback_storage
+    ) or Path("")
+    feedback_dir.mkdir(parents=True, exist_ok=True)
 
     current_time = str(datetime.now(UTC))
     data_to_store = {"user_id": user_id, "timestamp": current_time, **feedback}
 
     # stores feedback in a file under unique uuid
-    feedback_file_path = storage_path / f"{get_suid()}.json"
+    feedback_file_path = feedback_dir / f"{get_suid()}.json"
     with open(feedback_file_path, "w", encoding="utf-8") as feedback_file:
         json.dump(data_to_store, feedback_file)
 
--- a/src/app/endpoints/feedback.py
+++ b/src/app/endpoints/feedback.py
@@ -113,7 +113,7 @@ def store_feedback(user_id: str, feedback: dict) -> None:
     # Create feedback directory if it doesn't exist
     feedback_dir = Path(
         configuration.user_data_collection_configuration.feedback_storage
-    ) or Path("")
+    )
     feedback_dir.mkdir(parents=True, exist_ok=True)
 
     current_time = str(datetime.now(UTC))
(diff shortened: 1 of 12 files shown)


src/app/endpoints/feedback.py
```diff
@@ -109,19 +109,18 @@ def store_feedback(user_id: str, feedback: dict) -> None:
...
-    # at the same location.
-    storage_path = Path(
-        configuration.user_data_collection_configuration.feedback_storage or ""
-    )
-    storage_path.mkdir(parents=True, exist_ok=True)
+
+    # Create feedback directory if it doesn't exist
+    feedback_dir = Path(
+        configuration.user_data_collection_configuration.feedback_storage
+    ) or Path("")
```
onmete, line 116: AFAIK, `Path("")` was added because of performance testing. Is it still required (can't find Vishnu's GH handle-pinged on slack)?

src/app/endpoints/feedback.py
```diff
@@ -109,19 +109,18 @@ def store_feedback(user_id: str, feedback: dict) -> None:
...
+    feedback_dir.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
onmete: I'll update this - with the changes in PR, the `configuration.user_data_collection_configuration.feedback_storage` always exists.

src/models/config.py
```diff
@@ -130,21 +130,19 @@ class UserDataCollection(BaseModel):
     """User data collection configuration."""
 
     feedback_enabled: bool = False
-    feedback_storage: Optional[str] = None
     transcripts_enabled: bool = False
-    transcripts_storage: Optional[str] = None
+    user_data_dir: str = "user_data"
```
tisnik, line 134: CodeRabbit is right, `Path` is better there

src/models/config.py
```diff
@@ -130,21 +130,19 @@ class UserDataCollection(BaseModel):
...
-        """Check that storage_location is set when enabled."""
-        if self.feedback_enabled and self.feedback_storage is None:
-            raise ValueError("feedback_storage is required when feedback is enabled")
-        if self.transcripts_enabled and self.transcripts_storage is None:
-            raise ValueError(
-                "transcripts_storage is required when transcripts is enabled"
-            )
-        return self
+    @property
+    def feedback_storage(self) -> str:
```
tisnik, line 138: return `Path` as it's used later, no need to have just `str` properties

onmete: updated

jrobertboos:
Hi @onmete, Im not too sure why this PR is necessary. I understand what you are bringing up regarding that OLS depends upon a specific archive structure however I don't believe that reducing the configuration is necessary especially since with the existing configuration OLS can define a structure that fits its requirements while still allowing for as much configuration as possible which will potentially benefit other teams that use lightspeed core. 

I think that as of right now it would benefit lightspeed core to have an extensive of a configuration in order to accommodate all teams.

umago:
The patch does look good to me in terms of code, but I would like to see a reply to @jrobertboos question as well. I'm not up to date with all the data version and decisions taken so, so it's better for all teams to be aligned before we go ahead with it.

onmete:
We'll take a different approach. Apologies for "wasted" review time.

PR #347, opened 2025-08-05
Add database persistence layer and conversation tracking system

omertuc:
# Background / Context

When the user queries lightspeed-stack, they get a conversation ID in the response that they can later use to retrieve the conversation history and continue the conversation.

# Issue / Requirement / Reason for change

We've put the burden of remembering the conversation ID on the user, which is not ideal. The user (e.g. a UI) has to store the conversation ID somewhere, e.g. browser local storage, but this is not ideal because it can easily get lost. It also complicated the UI implementation.

Another issue is that a user could read another user's conversation if they only knew the ID, which is hard, but not ideal. It's not immediately obvious that the conversation ID is a secret that should not be shared with others.

# Solution / Feature Overview

Add a database persistence layer to lightspeed-stack to store user conversation IDs and a bit of metadata about the conversation.

Allow users to retrieve their conversations by calling a new endpoint `/conversations` (without an ID - with an ID is an existing endpoint that lists the message history for a conversation), which will return a list of conversations associated with the user, along with metadata like creation time, last message time, and message count.

Verify that the user owns the conversation when querying or streaming a conversation, and automatically associate the conversation with the user in the database when the user queries or streams a conversation.

# Implementation Details

**Database Layer (src/app/database.py):**
- SQLAlchemy
- Support for both SQLite and Postgre databases
- PostgreSQL custom schema (table namespacing) support with automatic
  schema creation. This is useful because llama-stack itself needs a
  database for persisting conversation history, and we want to avoid
  conflicts with that when users use the same Postgres database as
  llama-stack.
- Automatically enable SQL statement tracing when debug logging is
  enabled

**Models (src/models/):**
- Base model class for shared SQLAlchemy configuration
- UserConversation model with indexed user_id fields for efficient querying, automatic timestamp tracking (created_at, last_message_at) and message count tracking

**Configuration Updates:**
- Extended configuration system to support database connection settings

**API:**
- Conversations endpoint without an ID now returns a list of
  conversations for the authenticated user, along with metadata
- Query/streaming endpoint automatically associates conversations
  with the user in the database, and verifies ownership of conversations

**Dependencies:**
- Added SQLAlchemy dependency

# Future Work

This change does not include migrations, so any future changes to the databases require writing custom backwards compatibility code. We should probably add Alembic or something simpler in the near future.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement

- Related Issue #
- Closes #

MGMT-21300

## Testing

More tests will be added (and unit tests will be fixed) once I get some initial feedback on the implementation.


src/app/endpoints/streaming_query.py
```diff
@@ -424,7 +426,23 @@ async def streaming_query_endpoint_handler(
     llama_stack_config = configuration.llama_stack_configuration
     logger.info("LLama stack config: %s", llama_stack_config)
 
-    _user_id, _user_name, token = auth
+    user_id, _user_name, token = auth
+
+    # Validate conversation ownership if conversation_id is provided
+    if query_request.conversation_id:
+        if not validate_conversation_ownership(user_id, query_request.conversation_id):
```
eranco74, line 433: I like this!
Should resolve https://issues.redhat.com/browse/MGMT-21008

eranco74: I think it also indirectly addresses https://github.com/lightspeed-core/lightspeed-stack/pull/317#discussion_r2243110477

eranco74 (commented):
Nice!!

omertuc:
> LGTM
> 
> First TODO: please create JIRA ticket - automatic cleanup of old conversations from DB in order to keep the DB size under defined max value 

https://github.com/lightspeed-core/lightspeed-stack/pull/347#issuecomment-3159961392

>Second TODO: please create JIRA ticket to update documentation (README.md)


https://issues.redhat.com/browse/LCORE-494

src/app/endpoints/conversations.py
```diff
@@ -109,10 +141,63 @@ def simplify_session_data(session_data: dict) -> list[dict[str, Any]]:
...
+                ConversationDetails(
+                    conversation_id=conv.id,
+                    created_at=conv.created_at.isoformat() if conv.created_at else None,
+                    last_message_at=(
+                        conv.last_message_at.isoformat()
+                        if conv.last_message_at
+                        else None
+                    ),
+                    message_count=conv.message_count,
+                    model=conv.model,
```
carbonin, line 173: Can't the model change per request?
Also I don't see this in the `conversations_list_responses` example.

src/models/conversations.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class UserConversation(Base):  # pylint: disable=too-few-public-methods
+    """Model for storing user conversation metadata."""
+
+    __tablename__ = "user_conversation"
+
+    id: Mapped[str] = mapped_column(primary_key=True)
+    user_id: Mapped[str] = mapped_column(index=True)
+    model: Mapped[str] = mapped_column(index=True)
```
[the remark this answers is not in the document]
omertuc: Hmm yeah, I can also use that as a hint to pick up a model when the request lacks one. Will add that

src/app/endpoints/conversations.py
omertuc: Yeah, maybe I'll rename it to "last_model". It's good for auto-selection of the model when a conversation is restarted

matysek:
Merging, The E2E Tests check is broken and was fixed in: https://github.com/lightspeed-core/lightspeed-stack/pull/376

PR #368, opened 2025-08-08
DO NOT MERGE: discussion needed

bparees:
Use the llamastack identifier for the model id, not the provider's model id

## Description
see:https://llama-stack.readthedocs.io/en/latest/distributions/configuration.html#resources

```
What’s with the provider_model_id field? This is an identifier for the model inside the provider’s model catalog. Contrast it with model_id which is the identifier for the same model for Llama Stack’s purposes. For example, you may want to name “llama3.2:vision-11b” as “image_captioning_model” when you use it in your Stack interactions. When omitted, the server will set provider_model_id to be the same as model_id.
```

If we look at the LLS config api we have, for example:
```
models:
  - model_id: benbot
    provider_id: openai
    model_type: llm
    provider_model_id: gpt-4o-mini
```

which results in this in the model list:
```
    {
      "identifier": "benbot",
      "provider_resource_id": "gpt-4o-mini",
      "provider_id": "openai",
      "type": "model",
      "metadata": {},
      "model_type": "llm"
    },
```

Note that other prepopulated models look like:
```
    {
      "identifier": "openai/gpt-4-turbo",
      "provider_resource_id": "gpt-4-turbo",
      "provider_id": "openai",
      "type": "model",
      "metadata": {},
      "model_type": "llm"
    },
```

As far as LLS is concerned, the model it is aware of is identified by the "identifier" aka "model_id".  The "provider_model_id" should not be exposed to clients.

When i want to tell LCORE and LLS that i want to use the "benbot" model, i should be referring to it as "benbot", not "gpt-4o-mini".

if i want to use the prepopulated "gpt-4-turbo" model from openai, i should use the model_id "openai/gpt-4-turbo" with a provider_id of "openai"

This code change attemps to implement that by modifying the model lookup logic.

Note that this may impact clients that were previously using the "provider_resource_id" reported by LLS + LCORE models apis, when calling back into LCORE.  That is why this PR needs discussion.

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/src/app/endpoints/query.py
+++ b/src/app/endpoints/query.py
@@ -171,13 +171,14 @@ def select_model_and_provider_id(
                 },
             ) from e
 
-    llama_stack_model_id = f"{provider_id}/{model_id}"
     # Validate that the model_id and provider_id are in the available models
     logger.debug("Searching for model: %s, provider: %s", model_id, provider_id)
-    if not any(
-        m.identifier == llama_stack_model_id and m.provider_id == provider_id
-        for m in models
-    ):
+    
+    def check_model(m):
+        logger.debug("Available model - model_identifier: %s, provider_model_id: %s, provider_id: %s", m.identifier, m.provider_resource_id, m.provider_id)
+        return m.identifier == model_id and m.provider_id == provider_id
+    
+    if not any(check_model(m) for m in models):
         message = f"Model {model_id} from provider {provider_id} not found in available models"
         logger.error(message)
         raise HTTPException(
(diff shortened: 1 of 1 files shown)


bparees:
https://github.com/lightspeed-core/lightspeed-reference-ui/pull/6 is an example of changing the client logic to use the correct llamastack model identifier.

bparees:
github issue for the upstream llamastack behavior aspects of this:
https://github.com/meta-llama/llama-stack/issues/3116

bparees:
i've opened https://issues.redhat.com/browse/LCORE-600 to track resolving this behavior.

PR #385, opened 2025-08-12
Support configuration env variable replacement

omertuc:
We want to support environment variable replacement in the configuration similarly to how it is done in llama-stack.

Example use case / motivation:

The config, which since 3f7ed75e0e46766301dfbf622de3bfd0debb0736 can contain a password to a database, requires users to move their entire config to a Kubernetes secret, as it now contains sensitive information, rather than a configmap. If instead we supported environment variable replacement, users could keep their config in a configmap and just set the password in a secret as k8s allows you to load secret values as environment variables in pods.

Since users of lightspeed-stack are already familiar with the syntax for environment variable replacement from llama-stack, we can use the same function from llama-stack to handle this.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement

- Related Issue #
- Closes #

## Testing

Checked that it works. Since it's a borrowed function, testing seems unnecessary

## Summary by CodeRabbit

* **New Features**
  * Configuration files now support environment variable placeholders. You can reference variables (e.g., ${DB_PASSWORD}, ${API_URL}) in your YAML config, and they will be substituted from your environment at load time.
  * Works transparently with existing configurations—no changes to keys or structure required. This helps keep sensitive or environment-specific values out of your config files while preserving current behavior.

### DIFF
--- a/src/configuration.py
+++ b/src/configuration.py
@@ -3,6 +3,10 @@
 import logging
 from typing import Any, Optional
 
+# We want to support environment variable replacement in the configuration
+# similarly to how it is done in llama-stack, so we use their function directly
+from llama_stack.core.stack import replace_env_vars
+
 import yaml
 from models.config import (
     Configuration,
@@ -15,6 +19,7 @@ from models.config import (
     InferenceConfiguration,
     DatabaseConfiguration,
 )
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


maorfr (approved):
/lgtm

PR #434, opened 2025-08-21
Add customization profiles

Jdubrick:
## Description

In Road Core there was a way to have a profile that allowed you to have customized prompts, validation, etc. See: https://github.com/road-core/service/tree/main/ols/customize

This PR aims to add that same functionality, starting with prompts. It adds my teams profile (RHDH) and updates the way you can customize prompts, allowing you to choose a profile instead. 

Also this PR updates the documentation to outline this new way to add prompts and adds ample tests.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement

- Related Issue # https://issues.redhat.com/browse/RHDHPAI-976
- Closes # https://issues.redhat.com/browse/RHDHPAI-976

## Testing

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,14 +278,16 @@ For data export integration with Red Hat's Dataverse, see the [Data Export Integ
 
 ## System prompt
 
-   The service uses the, so called, system prompt to put the question into context before the question is sent to the selected LLM. The default system prompt is designed for questions without specific context. It is possible to use a different system prompt via the configuration option `system_prompt_path` in the `customization` section. That option must contain the path to the text file with the actual system prompt (can contain multiple lines). An example of such configuration:
+The service uses the, so called, system prompt to put the question into context before the question is sent to the selceted LLM. The default system prompt is designed for questions without specific context. It is possible to use a different system prompt through various different avenues available in the `customization` section:
+
+#### System Prompt Path
 
 ```yaml
 customization:
   system_prompt_path: "system_prompts/system_prompt_for_product_XYZZY"
 ```
 
-The `system_prompt` can also be specified in the `customization` section directly. For example:
+#### System Prompt Literal
 
 ```yaml
 customization:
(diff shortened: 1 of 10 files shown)


Jdubrick:
@tisnik in regards to hitting parity with how RCS was with the custom profiles (https://github.com/road-core/service/tree/main/ols/customize), to not use any customer-specific code, what would you recommend? Is it because the profile would be located in the codebase, would it be preferred if a path was provided instead?

manstis:
We, Ansible Lightspeed, have an interest in profiles or however the ability to switch between configurations at runtime is to be termed/defined/implemented. 

We have a use-case to support customers switching between, for example, WatsonX or vllm or Azure inference providers at runtime; together with different models. I.e. without having to rebuild `lightspeed-stack` related images. 

We know this will require the ability to define different System Prompts as, in our experience, the wording of it can differ between providers and models to optimise  the LLMs response and optimise Tool Call selection.

So, having some way to say "Use X profile" or "Use Y profile" etc would be very welcome. IIRC `ols` providers supported this type of function; although adding different System Prompts to the mix represents an enhancement over `ols`.

manstis:
@ldjebran @TamiTakamiya @jameswnl JFYI

https://github.com/lightspeed-core/lightspeed-stack/pull/434#issuecomment-3227236154

We can probably overload our container image and `llama-stack` _run_ config with dependencies and inference definitions to support customers swapping between different providers at runtime (like @TamiTakamiya added Gemeni dependencies already, but we have different _run_ configs for different providers rather than include them _all_ in a single file). We could handle different System Prompts for different providers by passing it in the `QueryRequest` but it'd be nicer if whatever `lightspeed-stack` is to offer (in the future) supports our need "out of the box".

This will relate to https://issues.redhat.com/browse/AAP-44959

manstis:
@tisnik 

> I think it should be added into LCORE

I wholeheartedly agree.. I'd prefer not to write project-specific code.

We use `Customization` at the moment for our System Prompt; however a new `Profiles` setting may be better suited.

Each _profile_ is in essence an amalgamation of:

- `Customization` (System Prompt)
- `InferenceConfiguration` (Model Id and Provider Id)

We have the additional consideration of possibly supporting different profile settings for external providers too:

- Question Validity [shield](https://github.com/lightspeed-core/lightspeed-providers/blob/main/lightspeed_stack_providers/providers/inline/safety/lightspeed_question_validity/config.py#L45).
- Custom [Agent](https://github.com/lightspeed-core/lightspeed-providers/blob/main/lightspeed_stack_providers/providers/inline/agents/lightspeed_inline_agent/config.py#L14) (tool call filtering)

I think profile support in the external providers is out of the initial scope.. it's not straight forward.

This PR [suggests](https://github.com/lightspeed-core/lightspeed-stack/pull/434/files#diff-138291983174f0a20f1aaf4c16782ccfc0dfe0ee0feda3e99ac429300b48a926R205-R216) @Jdubrick may have had similar thoughts; but it's not clear how they were intended to be used.

Jdubrick:
@tisnik it sounds like we want to keep this profile type of customization out of the project code so I can close this PR. While it seems like there is a desire for my team and others to have this kind of customization however, I know you mentioned via config file. Is it okay if I rework these changes and open a new PR where you pass a profile via a config file?

cc @yangcao77

PR #459, opened 2025-08-27
fix(auth): Use token in `AuthTuple`

omertuc:
While implementing authz I've made the authentication module put the claims in the `AuthTuple` instead of putting the original token because I assumed the original token was unnecessary and I didn't want to make the authz module re-parse the token to get the claims.

That was a mistake, because we need to pass on the user token to the MCP server for it to validate the user's access. So this commit changes the `AuthTuple` to contain the original token instead of the claims.

The authz module has been updated to parse the claims from the token instead of receiving them as a JSON string. Somewhat hackily, since we don't want to re-verify the token signature, so we assume that the token has already been verified during authentication and just decode the claims from the middle section of the JWT token.

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement

- Related Issue #
- Closes #

## Testing

Adjusted unit-tests, tested that MCP now works correctly manually

## Summary by CodeRabbit

- New Features
  - Added a JWT-based role resolver that derives user roles from token claims using configurable rules.
- Behavior Changes
  - Authentication now propagates the raw user token instead of parsed claims, while still validating user identity.
- Security
  - Introduced safer handling of JWT claims extraction for role evaluation.

### DIFF
--- a/src/auth/jwk_token.py
+++ b/src/auth/jwk_token.py
@@ -1,7 +1,6 @@
 """Manage authentication flow for FastAPI endpoints with JWK based JWT auth."""
 
 import logging
-import json
 from asyncio import Lock
 from typing import Any, Callable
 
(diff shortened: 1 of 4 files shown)


tisnik (approved):
LGTM

PR #475, opened 2025-09-01
allow disabling query model and provider

zszabo-rh:
## Description

Add tests to enforce and verify disablement of model/provider overrides via `customization.disable_query_model_provider_override`. Includes utils-level validation and endpoint-level behavior for `/query` and `/streaming_query`.

This ensure requests cannot override model/provider when the instance sets `customization.disable_query_model_provider_override: true`, and surface a clear 422 error with precise messaging.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement

- Related Issue [MGMT-21338](https://issues.redhat.com/browse/MGMT-21338)


tisnik (approved):
LGTM

omertuc:
> Should this simply be a `models.Action`, using the authorization system to give users more fine-grained control?

WDYT?

omertuc:
> > Should this simply be a `models.Action`, using the authorization system to give users more fine-grained control?
> 
> Unsure if that is correct, looking at the query endpoint implementation, I don't see where we enforce `models.Action` authorization.
> The models.Action is enforcing authz for the models endpoints, it's not doing anything for the query/streaming_query endpoints.
> Do you think we should change that?
> Note that overriding the models in the query is useful for evaluating different models.
> 
> 

I don't understand. I'm suggesting that this should be an action you can grant to a role rather than a global config for the entire deployment

zszabo-rh:
> > > Should this simply be a `models.Action`, using the authorization system to give users more fine-grained control?
> > 
> > 
> > Unsure if that is correct, looking at the query endpoint implementation, I don't see where we enforce `models.Action` authorization.
> > The models.Action is enforcing authz for the models endpoints, it's not doing anything for the query/streaming_query endpoints.
> > Do you think we should change that?
> > Note that overriding the models in the query is useful for evaluating different models.
> 
> I don't understand. I'm suggesting that this should be an action you can grant to a role rather than a global config for the entire deployment

I definitely see the pros here, though I also kinda like the simplicity of the current global config–based approach. The key question is whether the benefits of introducing RBAC would truly justify the added complexity.. do we expect strong need for it? 

My concern is mainly about implementing something that might add overhead without being widely adopted.

eranco74:
I don't mind one way or the other.

PR #476, opened 2025-09-01
Properly delete conversation

omertuc:
## Description

We saw that when deleting a conversation, it only gets deleted from the
llama-stack database, but not from the lightspeed-stack conversations
table.

This commit makes it that the conversation is deleted from both places,
ensuring it does not show up when the user lists conversations.

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement

- Related Issue #
- Closes #

MGMT-21651

## Testing

Unit tests have been adjusted. I also tested manually and saw the conversation no longer appearing after deletion.


tisnik:
> > don't we use `user_id` in a compound key?
> 
> What do you mean?

the user_id and conversation_id used to be a compound key in database. And because user_id can't be mocked (it's taken from the token) it means, that nobody else will be able to delete (or even read) user conversation, even if anybody tries to send bunch of random conversation_ids into the REST API.

omertuc:
> > > > > don't we use `user_id` in a compound key?
> > > > 
> > > > 
> > > > What do you mean?
> > > 
> > > 
> > > the user_id and conversation_id used to be a compound key in database. And because user_id can't be mocked (it's taken from the token) it means, that nobody else will be able to delete (or even read) user conversation, even if anybody tries to send bunch of random conversation_ids into the REST API.
> > 
> > 
> > They're not a compound primary key
> 
> does not it mean than anybody can access any conversation then?

No, we use `validate_conversation_ownership`

omertuc:
> > > > > > > don't we use `user_id` in a compound key?
> > > > > > 
> > > > > > 
> > > > > > What do you mean?
> > > > > 
> > > > > 
> > > > > the user_id and conversation_id used to be a compound key in database. And because user_id can't be mocked (it's taken from the token) it means, that nobody else will be able to delete (or even read) user conversation, even if anybody tries to send bunch of random conversation_ids into the REST API.
> > > > 
> > > > 
> > > > They're not a compound primary key
> > > 
> > > 
> > > does not it mean than anybody can access any conversation then?
> > 
> > 
> > No, we use `validate_conversation_ownership`
> 
> yup that's good. Also user_id is stored in db and conversation retrieved by:
> 
> ```
>         existing_conversation = (
>             session.query(UserConversation)
>             .filter_by(id=conversation_id, user_id=user_id)
>             .first()
>         )
> ```
> 
> so it looked like user_id+conversation_id are compound key. Whatever it look ok then

Ah I see. Yeah that's a bug, IMO this should only query by `conversation_id`. I'll do that in a separate PR

tisnik (approved):
LGTM

PR #1482, opened 2026-04-10
LCORE-1591: Observability for Lightspeed Core

asimurka:
## Description

Adds the feature architecture document for OpenTelemetry tracing in Lightspeed Core. Feature Specification only; no code  changes.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [x] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Assisted-by:  Cursor

- Related Issue # [LCORE-322](https://redhat.atlassian.net/browse/LCORE-322)
- Closes # [LCORE-1591](https://redhat.atlassian.net/browse/LCORE-1591)

## Testing

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive OpenTelemetry tracing design: global enablement and mandatory OTLP export config when enabled, inbound/outbound W3C trace-context propagation toggles, recommended instrumentation scope and span naming/attributes, lifecycle and resilience behavior (initialize/flush/gated operation), config precedence (hybrid/env-first guidance), deployment/packaging notes, privacy constraints, and an open question on library-mode limits.

### DIFF
--- a/docs/design/observability-opentelemetry/observability-opentelemetry.md
+++ b/docs/design/observability-opentelemetry/observability-opentelemetry.md
@@ -1,0 +1,637 @@
+# OpenTelemetry tracing in Lightspeed Core
+
+|                          |                                                                                   |
+|--------------------------|-----------------------------------------------------------------------------------|
+| **Date**                 | 2026-04-08                                                                        |
+| **Component**            | lightspeed-stack                                                                  |
+| **Authors**              | Andrej Šimurka                                                                    |
+| **Feature / Initiative** | [LCORE-322](https://redhat.atlassian.net/browse/LCORE-322)                        |
+| **Spike**                | [LCORE-1591](https://redhat.atlassian.net/browse/LCORE-1591)                      |
+| **Links**                | —                                                                                 |
+
+This document is the feature specification for observability in Lightspeed Core. It defines how to integrate comprehensive observability and tracing by leveraging the existing OpenTelemetry collector support in the upstream Llama Stack. It covers background, requirements, use cases, and architecture, including areas where multiple approaches are possible and which options are recommended.
+
+---
+
+## OpenTelemetry terminology
+
+- **Trace**: A complete record of a single request as it flows through one or more services. A trace is composed of multiple spans linked together via context propagation.
+
+- **Span**: A timed operation representing a unit of work within a trace (e.g., HTTP request handling, LLM call, RAG retrieval). Spans can be nested to reflect parent–child relationships.
+
+- **Attributes**: Key–value pairs attached to a span that describe its properties (e.g., model ID, token counts). Elapsed time for the operation is represented by the span’s own start/end, not duplicated as a duration attribute. Attributes should be low-cardinality and must not contain sensitive data.
+
+- **Events**: Timestamped annotations within a span that capture significant moments during execution (e.g., `stream.first_delta`, `llm.response.completed`). Events are not for bulk data, but for marking milestones.
+
+---
+
+## Background
+
+### Llama Stack
+
+Llama Stack supports **OpenTelemetry (OTel)** and can export traces and metrics via **OTLP** when configured through standard `OTEL_*` environment variables. 
+
+It provides in-process tracing and metrics via the OTel SDK. When deployed with standard HTTP instrumentation, it can **extract W3C trace context (`traceparent`, `tracestate`) from incoming requests**, allowing spans to attach to an upstream trace when context-providing headers are present.
+
+Configuration of telemetry in Llama Stack is controlled entirely by its own runtime configuration (`OTEL_*` environment variables) and is not managed or influenced by LCORE.
+
+---
+
+### Lightspeed Core
+
+LCORE exposes only Prometheus-compatible metrics via the `/metrics` endpoint. OpenTelemetry is not supported yet: there are no traces, spans, or OTLP metrics, and no configuration exists for enabling or controlling OTEL. All observability today relies entirely on Prometheus scraping.
+
+## What
+
+This feature introduces distributed tracing into LCORE using the OpenTelemetry Python SDK.
+
+It provides:
+
+- Configuration for tracing, including required OTLP export settings (endpoint, protocol, service name) and context propagation controls (incoming and outgoing)
+- Automatic HTTP server spans for the FastAPI application
+- Manual spans for key execution stages such as LLM calls, RAG processing, tool execution, moderation, and conversation management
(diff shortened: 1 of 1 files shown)


docs/design/observability-opentelemetry/observability-opentelemetry.md
```diff
@@ -0,0 +1,599 @@
...
+**Pros:** Sink basics are explicit and visible alongside other LCORE settings; advanced OTEL options remain on the standard environment path; avoids fully modeling OpenTelemetry in YAML.  
+**Cons:** Operators manage two surfaces; precedence rules between YAML and env vars must be clear.  
+
+**Normative precedence for Option 3:**  
+- If `enabled: false`, no TracerProvider or exporters are created; LCORE incurs no tracing overhead regardless of `OTEL_*` or stale YAML values.  
+- If `enabled: true`, YAML must contain mandatory **endpoint**, **protocol**, and **service name**; startup should fail if missing.  
+- When both YAML and env vars define the same concern, **YAML mandatory fields take precedence** for endpoint, protocol, and service name; `OTEL_*` variables control optional advanced settings (sampling, headers, TLS files, instrumentor-only options).  
+
+**Recommendation:** Option 1 is rejected for maintainability. Option 2 remains viable if LCORE is run with `opentelemetry-instrument`. Option 3 leverages LCORE’s configuration-driven design to ensure mandatory fields are always explicit when tracing is enabled.  
+
```
tisnik, line 179: option #2 seems promising.

Q - will those options be visible in config endpoint? If not, how we help admins with identifying tracing-related problems?

Q. for the future - how to handle configuration changes in this situation?

docs/design/observability-opentelemetry/observability-opentelemetry.md
```diff
@@ -0,0 +1,599 @@
...
+### Step 8: Prometheus Metrics Extension
+
+LCORE continues to expose **Prometheus-compatible metrics** via the existing `/metrics` endpoint using the native Prometheus Python SDK.  
+
+While OpenTelemetry tracing is introduced for spans, **metrics remain on Prometheus**, ensuring backward compatibility with scraping and alerting setups.  
+
+**Recommendation:**
+- Continue using the `/metrics` endpoint for all operational metrics.
+- Expand the set of Prometheus metrics as needed to cover additional components (e.g., LLM calls, RAG retrieval, tool execution) as product needs evolve.
```
tisnik, line 431: yes, it would be very valuable

docs/design/observability-opentelemetry/observability-opentelemetry.md
```diff
@@ -0,0 +1,599 @@
...
+- Ensure metrics align with existing naming conventions and maintain low cardinality to avoid high-memory cost in Prometheus servers.
```
tisnik, line 432: is it a real problem now? Or just warning/note for the future. Because now (IIRC) the cardinality selectors are just provider+model...

tisnik (commented):
It looks ok!

Q. for discussion: what happen if we won't include any Llama Stack traces? IMHO not much harm will be done as the only relevant + visible part is LCORE. In Llama Stack, the flow can change in future, so tracing will be a bit unstable from (for example) point of view of engineer, who wants to improve performance.

Context: almost all LS calls are pretty low level, so even LCORE traces might have enough granularity.

But I'm open for discussion.

docs/design/observability-opentelemetry/observability-opentelemetry.md
asimurka: No, these options won't be visible in config endpoint by default. We can possibly implement some "injection" into `v1/config` response model if desired. Problem is that there is a lot of configurable options for OTEL (see this [link](https://opentelemetry-python.readthedocs.io/en/latest/sdk/environment_variables.html)) but it is possible to collect all environment variables with `OTEL_` prefix and inject them into config.

To the best of my knowledge, `OTEL_*` environment variables are collected during process startup. As a result, any change to these variables requires the service process to be restarted/redeployed to apply changes.

docs/design/observability-opentelemetry/observability-opentelemetry.md
asimurka: I think it isn't a real problem rn.

asimurka:
> Q. for discussion: what happen if we won't include any Llama Stack traces?

I have an impression that the general expectation in distributed systems is that when you have multi-service integration, spans from downstream services should be part of the same trace to preserve end-to-end observability. From what I’ve seen, Llama Stack already provides fairly rich and detailed tracing, and it would be valuable to leverage that (at least in server mode) rather than losing that level of visibility.

Even with LLS spans, LCORE traces would remain complete and consistent on their own, but LLS spans would provide optional deeper insight into request-level behavior (especially latency breakdowns for more complex calls).

asimurka:
> almost all LS calls are pretty low level

Tbh, I don't think so. From what I can see in Jaeger, Llama Stack traces are pretty granular, with a lot of internal spans exposed per request.

tisnik:
> > almost all LS calls are pretty low level
> 
> Tbh, I don't think so. From what I can see in Jaeger, Llama Stack traces are pretty granular, with a lot of internal spans exposed per request.

I agree, but my point is who will benefit from it. Admin/devops who look for perf. problems wont be probably interested in LS internals. More interesting is (IMHO) "is the RAG what caused the slowdown" or "which MCP call is the most time expensive?" etc. OTOH we migth be interested knowing tracing about LS providers, that would be good, indeed.

asimurka:
> More interesting is (IMHO) "is the RAG what caused the slowdown" or "which MCP call is the most time expensive?" etc. OTOH we migth be interested knowing tracing about LS providers, that would be good, indeed.

Yeah, I agree but exactly for this we need LLS traces, right? Otherwise this would be completely hidden under a single LCORE responses request span.

jrobertboos:
@asimurka just wondering why the spike document is not included in this PR?

asimurka:
> @asimurka just wondering why the spike document is not included in this PR?

The document I propose IMO contains everything necessary for scoping the feature, including decision points.

docs/design/observability-opentelemetry/observability-opentelemetry.md
tisnik: @coderabbitai create issue

tisnik (approved):
LGTM

PR #1484, opened 2026-04-10
LCORE-1594: Agent Skills feature breakdown

jrobertboos:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Related Issue #
- Closes #

## Testing

## Summary by CodeRabbit

* **Documentation**
  * Added design and spike docs for Agent Skills: defines a configuration-driven skills catalog, required skill metadata and instruction files, startup validation and error responses, tools for listing/activating/loading skill content and resources, path-restricted resource access, protections for injected skill content, per-conversation activation deduplication, deferral of script execution, plus examples, tests, and integration guidance.

### DIFF
--- a/docs/design/agent-skills/agent-skills-spike.md
+++ b/docs/design/agent-skills/agent-skills-spike.md
@@ -1,0 +1,361 @@
+# Spike for Agent Skills Support
+
+## Overview
+
+**The problem**: Lightspeed Core has no mechanism for extending agent capabilities with specialized instructions or domain knowledge. Users cannot package reusable workflows, troubleshooting guides, or domain expertise into portable, discoverable units that the LLM can use on demand.
+
+**The recommendation**: Implement the [Agent Skills open standard](https://agentskills.io) with a config-based discovery model. Skills are defined in `lightspeed-stack.yaml` pointing to directories containing `SKILL.md` files. The LLM sees a skill catalog (name + description) in the system prompt and can request full instructions on demand via a dedicated tool.
+
+**PoC validation**: Not applicable for this spike. The feature is well-defined by the agentskills.io specification and has been implemented by 30+ agent products including Claude Code, GitHub Copilot, Cursor, and OpenAI Codex.
+
+## Decisions for @sbunciak and @ptisnovs
+
+These are the high-level decisions that determine scope, approach, and cost. Each has a recommendation confirmed during spike research.
+
+### Decision 1: Which skill types to support?
+
+| Option | Description |
+|--------|-------------|
+| A | Built-in skills only (Lightspeed developers ship pre-defined skills) |
+| B | Customer-defined only (end users import their own skill definitions) |
+| C | Both built-in and customer-defined |
+
+**Recommendation**: **B** (Customer-defined only). This allows end users to extend Lightspeed with their own domain expertise without requiring changes to the core product. Built-in skills can be added later if needed.
+
+### Decision 2: Discovery mechanism?
+
+| Option | Description |
+|--------|-------------|
+| A | Filesystem-based (scan configured directories for `SKILL.md` files) |
+| B | Config-based (define skills in `lightspeed-stack.yaml`) |
+| C | API-based (skills registered/managed via REST API) |
+| D | Hybrid (built-in via config, customer-defined via filesystem or API) |
+
+**Recommendation**: **B** (Config-based). Skills are defined in `lightspeed-stack.yaml` similar to `mcp_servers`. This provides explicit control over which skills are available, integrates with existing configuration patterns, and avoids filesystem scanning complexity.
+
+### Decision 3: Script execution support?
+
+| Option | Description |
+|--------|-------------|
+| A | No scripts (only `SKILL.md` instructions) |
+| B | Scripts allowed (full spec compliance) |
+| C | Deferred (start with no scripts, add later) |
+
+**Recommendation**: **A** (No scripts). As noted in LCORE-1339, there are security concerns with executing arbitrary scripts. The core value of skills is in the instructions — scripts can be added in a future phase after security review if needed.
+
+## Technical decisions for @ptisnovs
+
+Architecture-level and implementation-level decisions.
+
+### Decision 4: Skill content source
+
+| Option | Description |
+|--------|-------------|
+| A | Path-based (config points to directory containing `SKILL.md`) |
+| B | Inline (instructions embedded directly in YAML) |
+| C | Both (support either path or inline content) |
+
+**Recommendation**: **A** (Path-based). This follows the agentskills.io spec directory structure, keeps the YAML config clean, and allows skills to include `references/` subdirectories for additional documentation that can be loaded on demand.
+
+### Decision 5: Activation mechanism
+
+| Option | Description |
+|--------|-------------|
+| A | System prompt injection (skill catalog in system prompt, LLM decides) |
+| B | Dedicated tool (register `activate_skill` tool, LLM calls to load) |
+| C | Per-request parameter (client specifies which skills to activate) |
+
+**Recommendation**: **A** (System prompt injection) for the catalog, combined with **B** (dedicated tool) for loading full instructions. The system prompt contains the skill catalog (name + description, ~50-100 tokens per skill). When the LLM decides a skill is relevant, it calls the `activate_skill` tool to load the full instructions.
+
+### Decision 6: Skill context management
+
+| Option | Description |
+|--------|-------------|
+| A | Always loaded (all skills' full instructions in every request) |
+| B | Catalog + on-demand (only name/description upfront, full content loaded when LLM requests) |
+
+**Recommendation**: **B** (Catalog + on-demand). This follows the agentskills.io progressive disclosure pattern:
+1. **Catalog** (~50-100 tokens/skill) - name + description always in system prompt
+2. **Instructions** (<5000 tokens) - full `SKILL.md` body loaded via tool when needed
+3. **Resources** (on-demand) - `references/` files loaded via file-read tool when referenced
+
+This keeps the base context small while giving the LLM access to specialized knowledge on demand.
+
+### Decision 7: Configuration structure
+
+Skills are configured as a list in `lightspeed-stack.yaml`, following the `mcp_servers` pattern:
+
+```yaml
+skills:
+  - name: "code-review"
+    description: "Review code for best practices and security issues."
+    path: "/var/skills/code-review"
+  - name: "troubleshooting"
+    description: "Diagnose and fix OpenShift deployment issues."
+    path: "/var/skills/troubleshooting"
+```
+
+Each skill entry specifies:
+- `name`: Unique identifier (validated against `SKILL.md` frontmatter)
+- `description`: What the skill does and when to use it
+- `path`: Absolute path to directory containing `SKILL.md`
+
+**Recommendation**: Approved. This structure is consistent with existing config patterns and provides explicit control.
+
+## Proposed JIRAs
+
+Each JIRA includes an agentic tool instruction pointing to the spec doc.
+
+<!-- type: Task -->
+<!-- key: LCORE-???? -->
+### LCORE-???? Add skill configuration model
+
+**Description**: Add the `SkillConfiguration` Pydantic model and `skills` list to the main configuration. This enables defining skills in `lightspeed-stack.yaml`.
+
+**Scope**:
+
+- Create `SkillConfiguration` class in `src/models/config.py`
+- Add `skills: list[SkillConfiguration]` field to `Configuration` class
+- Add validation: path exists, contains `SKILL.md`, name matches frontmatter
+- Parse `SKILL.md` frontmatter on startup (extract name, description)
+
+**Acceptance criteria**:
+
+- Skills can be defined in `lightspeed-stack.yaml` using the documented format
+- Startup fails with clear error if skill path doesn't exist or lacks `SKILL.md`
+- Startup fails if configured `name` doesn't match `SKILL.md` frontmatter `name`
+- Unit tests cover validation scenarios
+
+**Agentic tool instruction**:
+
+```text
+Read the "Configuration" section in docs/design/agent-skills/agent-skills.md.
+Key files: src/models/config.py (around line 1817, Configuration class).
+Follow the MCP server config pattern (ModelContextProtocolServer class, line 468).
+```
+
+<!-- type: Task -->
+<!-- key: LCORE-???? -->
+### LCORE-???? Implement skill catalog injection
+
+**Description**: Inject the skill catalog (name + description for each skill) into the system prompt so the LLM knows what skills are available.
+
+**Scope**:
+
+- Create `src/utils/skills.py` module
+- Implement `build_skill_catalog()` function that formats skills as XML/structured text
+- Modify `get_system_prompt()` in `src/utils/prompts.py` to append skill catalog
+- Add behavioral instructions telling the LLM how to use skills
+
+**Acceptance criteria**:
+
+- System prompt includes skill catalog when skills are configured
+- Catalog format includes name, description, and path for each skill
+- Catalog is omitted when no skills are configured
+- Unit tests verify catalog formatting and system prompt injection
+
+**Agentic tool instruction**:
+
+```text
+Read the "System prompt injection" section in docs/design/agent-skills/agent-skills.md.
+Key files: src/utils/prompts.py (get_system_prompt function),
+src/utils/skills.py (new module).
+```
+
+<!-- type: Task -->
+<!-- key: LCORE-???? -->
+### LCORE-???? Implement activate_skill tool
+
+**Description**: Register a `activate_skill` tool that the LLM can call to load full skill instructions when it decides a skill is relevant.
+
+**Scope**:
+
+- Add `activate_skill` tool registration in `prepare_tools()` in `src/utils/responses.py`
+- Implement tool handler that reads `SKILL.md` body content
+- Return structured response with skill content and base directory path
+- Optionally list `references/` files if present
+
+**Acceptance criteria**:
+
+- LLM can call `activate_skill(name="skill-name")` to load skill instructions
+- Tool returns full `SKILL.md` body content (after frontmatter)
+- Tool returns skill directory path so LLM can resolve relative references
+- Tool returns error if skill name is invalid
+- Unit tests cover tool registration and invocation
+
+**Agentic tool instruction**:
+
+```text
+Read the "Skill activation tool" section in docs/design/agent-skills/agent-skills.md.
+Key files: src/utils/responses.py (prepare_tools function, line 204),
+src/utils/skills.py.
+```
+
+<!-- type: Task -->
+<!-- key: LCORE-???? -->
+### LCORE-???? Add skill reference file access
+
+**Description**: Enable the LLM to read files from the skill's `references/` subdirectory when skill instructions reference them.
+
+**Scope**:
+
+- Ensure existing file-read tool can access skill reference directories
+- Add path validation to restrict access to configured skill directories only
+- Document the pattern for skill authors to reference files
+
+**Acceptance criteria**:
+
+- LLM can read files from `<skill-path>/references/` using existing file-read capability
+- Access is restricted to configured skill directories (no arbitrary filesystem access)
+- Skill instructions can use relative paths like `references/troubleshooting-guide.md`
+- Integration test verifies reference file access works end-to-end
+
+**Agentic tool instruction**:
+
+```text
+Read the "Reference file access" section in docs/design/agent-skills/agent-skills.md.
+Key files: src/utils/responses.py, existing file-read tool implementation.
+```
+
+<!-- type: Story -->
+<!-- key: LCORE-???? -->
+### LCORE-???? Document Agent Skills feature
+
+**Description**: Create user-facing documentation for the Agent Skills feature including configuration guide, skill authoring guide, and examples.
+
+**Scope**:
+
+- Add configuration documentation to existing config docs
+- Create skill authoring guide (SKILL.md format, directory structure)
+- Add example skills to `examples/skills/`
+- Update README with feature overview
+
+**Acceptance criteria**:
+
+- Users can configure skills by following the documentation
+- Skill authors can create compliant skills using the authoring guide
+- Example skills demonstrate common use cases (troubleshooting, code review, etc.)
+
+**Agentic tool instruction**:
+
+```text
+Read the full spec doc at docs/design/agent-skills/agent-skills.md.
+Reference the agentskills.io specification for SKILL.md format details.
+```
+
+<!-- type: Task -->
+<!-- key: LCORE-???? -->
+### LCORE-???? Add integration and E2E tests for skills
+
+**Description**: Add integration tests and E2E tests to verify the skills feature works correctly end-to-end.
+
+**Scope**:
+
+- Integration tests: skill loading, catalog injection, tool invocation with mocked LLM
+- E2E tests: full flow with real LLM, verify skill activation and usage
+
+**Acceptance criteria**:
+
+- Integration tests cover skill configuration, catalog generation, and tool handling
+- E2E tests verify a configured skill can be discovered and used by the LLM
+- Tests use example skills from `examples/skills/`
+
+**Agentic tool instruction**:
+
+```text
+Read the "Testing" section in docs/design/agent-skills/agent-skills.md.
+Key test files: tests/integration/endpoints/, tests/e2e/features/.
+Follow existing test patterns in the codebase.
+```
+
+## Background sections
+
+### Agent Skills specification
+
+The [Agent Skills open standard](https://agentskills.io) defines a portable format for giving AI agents specialized capabilities. Key elements:
+
+**Directory structure**:
+```
+skill-name/
+├── SKILL.md          # Required: metadata + instructions
+├── references/       # Optional: additional documentation
+└── assets/           # Optional: templates, resources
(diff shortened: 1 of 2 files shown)


docs/design/agent-skills/agent-skills-spike.md
```diff
@@ -0,0 +1,361 @@
+# Spike for Agent Skills Support
+
+## Overview
+
+**The problem**: Lightspeed Core has no mechanism for extending agent capabilities with specialized instructions or domain knowledge. Users cannot package reusable workflows, troubleshooting guides, or domain expertise into portable, discoverable units that the LLM can use on demand.
+
+**The recommendation**: Implement the [Agent Skills open standard](https://agentskills.io) with a config-based discovery model. Skills are defined in `lightspeed-stack.yaml` pointing to directories containing `SKILL.md` files. The LLM sees a skill catalog (name + description) in the system prompt and can request full instructions on demand via a dedicated tool.
```
anik120: I've noticed this "pointing to directories" pattern throughout this project, but want to potentially explore a new way via this feature if possible - 

If we look to the Kubernetes world to guide us in this for example, helm charts can be found in https://artifacthub.io, and some helm charts are distributed using container registries, some distributed using GitHub, and some using standalone websites. Here's a popular helm chart, https://artifacthub.io/packages/helm/percona/ps-db,  distributed using GitHub  - https://percona.github.io/percona-helm-charts/ 

Notice that the way for me to use the ps-db helm chart artifacts (which are files in a folder), is not "clone [helm GitHub repository](https://github.com/helm/helm), use this directory" - that wouldn't be scalable once it got to 100+ helm charts. 

Which means it won't be scalable for us when we get 10+ agent skill contribution. 

Here's examples of the three scenarios if we were to use them for distributing agent skills: 

1. using container registries - https://docker.io/anik120/markdown-to-gdoc-converter-agent-skill
2. using GitHub - https://anik120.github.io/markdown-to-gdoc-converter-agent-skill/
3. Using standalone website - https://my-personal-website/markdown-to-gdoc-converter-agent-skill


I think for our use case, the best first step is to start with a GitHub distribution mechanism. That would mean we'd need to build in mechanism to use the GitHub apis to pull down these artifacts (and make sure they contain artifacts in the filesystem that conform to the standards put forth in agentskills.io). 


This will help Red Hat productize agent skills too - we can have Red hat owned repositories that contain these skills, that are submitted by partner eco-system partners (and others), but vetted by Red Hat. 


...the same pattern can then be developed for `providers" and other stuff...but we can start with agent skills as the implementation guidance for other artifacts.

docs/design/agent-skills/agent-skills-spike.md
```diff
@@ -0,0 +1,361 @@
...
+
+### Decision 3: Script execution support?
+
+| Option | Description |
+|--------|-------------|
+| A | No scripts (only `SKILL.md` instructions) |
+| B | Scripts allowed (full spec compliance) |
+| C | Deferred (start with no scripts, add later) |
+
+**Recommendation**: **A** (No scripts). As noted in LCORE-1339, there are security concerns with executing arbitrary scripts. The core value of skills is in the instructions — scripts can be added in a future phase after security review if needed.
```
anik120: If we adopt the agent skill distribution I suggested above, we can have script execution capabilities - Red Hat verified agent skills (with full, script execution capabilities) in a Red Hat owned repo would be the business value we bring - if you're an enterprise customer using the agent skills from the Red Hat owned repos, you can be sure your system won't be compromised with script execution.

docs/design/agent-skills/agent-skills-spike.md
anik120: ie to use the example you have below, we'd modify it to look like: 

```yaml
skills:
  - name: "code-review"
    description: "Review code for best practices and security issues."
    path: https://redhat.github.io/code-review-agent/
  - name: "troubleshooting"
    description: "Diagnose and fix OpenShift deployment issues."
    path: https://redhat.github.io/openshift-troubleshooter-agent/
```

bparees: pulling skills down from external urls introduces significant security/trust concerns.  Not only "do i trust this url today" but "what happens if the content at that url is compromised or updated in the future?"

For that reason alone I think starting w/ reading from the filesystem is correct.  

But in addition, there are many ways, within k8s, to inject content into a container filesystem.  So there's no reason that in the future, you can't add adapters that pull container from github, or copy content over from a sidecar container or uses image volumes to pull content from a "skills library" image, or even define skills in configmaps that get mounted in.

So my recommendation would be to start by assuming LCORE gets its skills from the filesystem.  Then later you can worry about the best ways to get those skills onto the filesystem from various marketplace/distribution sources.

docs/design/agent-skills/agent-skills-spike.md
```diff
@@ -0,0 +1,361 @@
...
+
+### Decision 1: Which skill types to support?
+
+| Option | Description |
+|--------|-------------|
+| A | Built-in skills only (Lightspeed developers ship pre-defined skills) |
+| B | Customer-defined only (end users import their own skill definitions) |
+| C | Both built-in and customer-defined |
+
+**Recommendation**: **B** (Customer-defined only). This allows end users to extend Lightspeed with their own domain expertise without requiring changes to the core product. Built-in skills can be added later if needed.
```
bparees: This seems to propose that the only way stack products (e.g. rhel lightspeed) would be able to ship skills would be if they were a part of(shipped with) lightspeed core?  Is that the intended statement?

It seems like the primary value of this feature would be to allow product teams to inject better agentic guidance into LCORE that serves their specific intended use cases.  And LCORE should enable that in a way that does not require those skills be packaged with LCORE itself.  (Such as by allowing skills to be injected into the LCORE filesystem via configmaps or container volumes provided by product teams)

customers adding their own skills also makes sense, but opens product teams up to the risk that a customer adds a skill that conflicts w/ their own logic, for better or worse.  So i'd expect a lot of careful thought to go into what it means to allow customers to do that.  It's like allowing them to edit the system prompt, effectively.

docs/design/agent-skills/agent-skills-spike.md
```diff
@@ -0,0 +1,361 @@
...
+### Decision 2: Discovery mechanism?
+
+| Option | Description |
+|--------|-------------|
+| A | Filesystem-based (scan configured directories for `SKILL.md` files) |
+| B | Config-based (define skills in `lightspeed-stack.yaml`) |
+| C | API-based (skills registered/managed via REST API) |
+| D | Hybrid (built-in via config, customer-defined via filesystem or API) |
+
+**Recommendation**: **B** (Config-based). Skills are defined in `lightspeed-stack.yaml` similar to `mcp_servers`. This provides explicit control over which skills are available, integrates with existing configuration patterns, and avoids filesystem scanning complexity.
```
bparees: So the skill definitions would be inlined in the lightspeed-stack.yaml?  I think that is a mistake.  For one thing, it means the config CR that drives LCORE needs to contain those skills, which potentially makes the CR huge (unless you're going to stitch the lightspeed-stack.yaml together from a config CR plus various Skill CRs).

It also makes the yaml file huge and hard to update when one particular skill needs to be modified.

MCP servers are not an equivalent config concept here because they just have a few well specified config fields that need to be represented.

Loading skills from the filesystem should not be complicated.  Getting the skill files *onto* the filesystem could be complicated, but again k8s offers numerous ways to accomplish this (and for non-k8s deployments, there are also many ways to solve this problem, none of them particularly novel or complicated)

docs/design/agent-skills/agent-skills-spike.md
bparees: until LCORE provides a sandbox mechanism to run said scripts in an ephemeral, isolated environment, I wholeheartedly agree w/ not supporting script execution.

jrobertboos: I like the idea of supporting more than just local paths for skills however it would be just as easy to execute scripts from a local directory the only reason I am recommending not implementing that functionality is according to the [feature ticket](https://redhat.atlassian.net/browse/LCORE-1339) it is out of scope.

docs/design/agent-skills/agent-skills-spike.md
```diff
@@ -0,0 +1,361 @@
...
+
+### Decision 4: Skill content source
+
+| Option | Description |
+|--------|-------------|
+| A | Path-based (config points to directory containing `SKILL.md`) |
+| B | Inline (instructions embedded directly in YAML) |
+| C | Both (support either path or inline content) |
+
+**Recommendation**: **A** (Path-based). This follows the agentskills.io spec directory structure, keeps the YAML config clean, and allows skills to include `references/` subdirectories for additional documentation that can be loaded on demand.
```
bparees, line 60: per my other comments, i'm +1 on path-based.

docs/design/agent-skills/agent-skills-spike.md
jrobertboos: I like this idea however I believe that this is something that could be added in a later feature as there would not be much architectural change in the system wether it be a local "directory"/path or a remote "directory"/path. I do think that a local directory is **required** while the remote aspect is a "_would be nice_" tho.

docs/design/agent-skills/agent-skills-spike.md
```diff
@@ -0,0 +1,361 @@
...
+
+### Decision 5: Activation mechanism
+
+| Option | Description |
+|--------|-------------|
+| A | System prompt injection (skill catalog in system prompt, LLM decides) |
+| B | Dedicated tool (register `activate_skill` tool, LLM calls to load) |
+| C | Per-request parameter (client specifies which skills to activate) |
+
+**Recommendation**: **A** (System prompt injection) for the catalog, combined with **B** (dedicated tool) for loading full instructions. The system prompt contains the skill catalog (name + description, ~50-100 tokens per skill). When the LLM decides a skill is relevant, it calls the `activate_skill` tool to load the full instructions.
```
bparees: I think this is ok, but i'm going to offer up an alternative approach:  Don't even include the skill descriptions in the system prompt.  Just give the llm a "skill_search" tool.  Let it invoke the skill_search tool to try to identify skills that may match the intent, and then it can pick from the search results to invoke activate_skill.  (The skill searching would be some sort of keyword+semantic hybrid search)

This might be a "phase 2" implementation.  The risk is that searching is ineffective in finding the right skills, but the upside is you don't risk overwhelming the model with too many skill descriptions in the system prompt.

Maybe consider a hybrid approach (as long as their are less than X skills, just include them all in the system prompt. When we get beyond X skills, move to a skill search approach.

docs/design/agent-skills/agent-skills-spike.md
jrobertboos: I can add somewhere tho to make the path handled by some sort of interface so we can have local **and** remote handlers for the skill directories.

docs/design/agent-skills/agent-skills-spike.md
```diff
@@ -0,0 +1,361 @@
...
+    description: "Diagnose and fix OpenShift deployment issues."
+    path: "/var/skills/troubleshooting"
+```
+
+Each skill entry specifies:
+- `name`: Unique identifier (validated against `SKILL.md` frontmatter)
+- `description`: What the skill does and when to use it
+- `path`: Absolute path to directory containing `SKILL.md`
+
+**Recommendation**: Approved. This structure is consistent with existing config patterns and provides explicit control.
```
bparees: Aren't skill.md files self-describing?  What is the benefit to explicitly configure individual skills vs just pointing to a directory of skills, the same way claude/cursor/etc do? (skills directory, containing a directory per skill)

I can envision a few benefits to this approach, but it also adds significant config management overheard and opportunity for mistakes (what happens when the path is incorrect?  what happens when the name/description don't match the skill.md file the path points to?)

docs/design/agent-skills/agent-skills.md
```diff
@@ -0,0 +1,644 @@
...
+
+### Context management
+
+Once skill instructions are in the conversation context, they must remain effective for the session duration.
+
+#### Protect skill content from compaction
+
+If lightspeed-stack implements context compaction (conversation history summarization), skill content must be exempted from pruning. Skill instructions are durable behavioral guidance — losing them mid-conversation silently degrades performance.
+
+The `<skill_content>` tags from structured wrapping enable identification during compaction:
```
bparees: this seems potentially fragile.  what happens if this string shows up in the context for other reasons?

Also at some point you may be unable to compact if you're unwilling to compact skill content.

docs/design/agent-skills/agent-skills.md
```diff
@@ -0,0 +1,644 @@
...
+
+```python
+class SkillTracker:
+    """Track activated skills per conversation."""
+
+    def __init__(self):
+        self._activated: dict[str, set[str]] = {}  # conversation_id -> skill names
+
+    def is_activated(self, conversation_id: str, skill_name: str) -> bool:
+        return skill_name in self._activated.get(conversation_id, set())
```
bparees, line 578: is this persisted somewhere?  if not, what are the implications for persisting+resuming conversations?

docs/design/agent-skills/agent-skills.md
```diff
@@ -0,0 +1,644 @@
...
+    return f"<note>Skill '{name}' is already loaded in this conversation.</note>"
+```
+
+### Error handling
+
+| Scenario | Error |
+|----------|-------|
+| Skill path doesn't exist | Startup fails: "Skill path does not exist: {path}" |
+| SKILL.md not found | Startup fails: "SKILL.md not found at {path}/SKILL.md" |
+| Name mismatch | Startup fails: "Skill name mismatch: config has '{x}' but SKILL.md has '{y}'" |
```
bparees, line 602: per my earlier comment, you could avoid this entire class of error by simplifying the config.yaml to not include this information.

docs/design/agent-skills/agent-skills-spike.md
anik120: Okay I think that makes sense. with 

>So there's no reason that in the future, you can't add adapters that pull container from github, or copy content over from a sidecar container or uses image volumes to pull content from a "skills library" image, or even define skills in configmaps that get mounted in.

we don't have to do 

>I can add somewhere tho to make the path handled by some sort of interface so we can have local and remote handlers for the skill directories.

this even. It can just be a plain simple path in the filesystem that is pointed to at the configuration - the abstraction for how the content is pulled then is taken care at the sidecar level

docs/design/agent-skills/agent-skills-spike.md
jrobertboos: so by customer I mean an LS app team (e.g. rhel lightspeed) and they should be able to have/maintain their own skills which then would host the skills in the container alongside lightspeed stack and would point using the lightspeed-stack configuration to the included skills. 

The customers (non-red hat) of the LS app teams would not have the ability to add skills just like they do not have the ability to add mcp servers.

bparees: ok in that case i think we are in agreement but i'd suggest a terminology cleanup because "customers" and "end users" is easy to misunderstand.

anik120: I read "Customers" as Openshift/RHEL/Ansible users too, might be worth clarifying this as "product team" (or something that identifies Openshift/RHEL/Ansible teams clearly)

docs/design/agent-skills/agent-skills-spike.md
jrobertboos: I agree with you that skill definitions are not necessarily needed to be defined in the config (I will update that tomorrow) as they can be fetched from the actual files. The only reason I did it this way was because I was thinking of the config being like the [skills catalog](https://agentskills.io/client-implementation/adding-skills-support#what-to-store). 

I am not sure though why the skills should not be similar to the MCP config tho at least in the sense that at the bare minimum we need a path.

And it could be that we supply just one path or a list of paths and its expected that that path leads to a directory of skills e.g.

```
~/.agents/skills/
├── pdf-processing/
│   ├── SKILL.md          ← discovered
│   └── scripts/
│       └── extract.py
├── data-analysis/
│   └── SKILL.md          ← discovered
└── README.md             ← ignored (not a skill directory)
```

p.s. sorry for the broken English, I am tired haha

docs/design/agent-skills/agent-skills-spike.md
jrobertboos: so the main reason that I did it this way by adding the skills to the context is because of [this](https://agentskills.io/client-implementation/adding-skills-support#the-core-principle-progressive-disclosure). I guess this could work well especially once a lot of skills get added but ur right in that I think that this would be a "phase 2" thing, I might change PR tho so that instead of appending to context a required `skill_search` tool that "for now" does not search but just returns the skill catalog list (names + descriptions). that would make extending the "search" part of the tool so that the model is not overwhelmed by large number of skills easy in the future.

docs/design/agent-skills/agent-skills.md
jrobertboos: So the main reason I did this is its the [recommendation](https://agentskills.io/client-implementation/adding-skills-support#structured-wrapping) from the agent skills spec. also I did not think about the compaction problem but I think that would only really be a problem if there were ton and TONS of skills, which I think we could cover in a "phase 2".

"Phase 2" could basically be increasing scalability to allow for integration of more than 100 (or any limiting number) skills. As this would cover working on the `skill_search` which would be needed for large numbers of skills and how to handle compaction with a very large skill catalog (I think that this is less of an issue because I am not finding anyone having problems with this online, but still it would be something to safeguard against).

also the reason we avoid compaction is because its defined in the spec [here](https://agentskills.io/client-implementation/adding-skills-support#protect-skill-content-from-context-compaction).

docs/design/agent-skills/agent-skills-spike.md
bparees: > I am not sure though why the skills should not be similar to the MCP config tho at least in the sense that at the bare minimum we need a path.

no disagreement there, my "this is not the same as mcp config" was only intended to mean "we don't need the full config definition inlined".  Having the skill config consist entirely of either a path per skill, or just a path to a directory of all the skills, is what i had in mind.

jrobertboos: ahhhh, ok that makes sense :)

Akrog: If configured path leads to a directory of skills and lightsped-stack recursively crawls all sub-directories looking for `SKILL.md` files then we are limiting things to a flat structure where all skills are at the top level, right?

I think that's an unnecessary limitation, since a nice feature of skills for complex systems is the possibility of nested skills.

And if we limit the auto-discovery of skills from the configured path to the first level of sub-directories, then we are also making it difficult to reuse some skills in multiple first-level skills.

Personally I think having the `path` configuration point to a directory with a `SKILL.md` file is the basic functionality to have.  If we want we could also have a `recursive_path` or something like that as well that auto-discovers skills.

docs/design/agent-skills/agent-skills-spike.md
Akrog: I was thinking of not opening skill support in OpenStack Lightspeed to users (at least initially), it would be a mechanism used by the Red Hat team to be able to add our skills to our product, so the code in `scripts` would be:
- Created by Red Hat
- Delivered and Configured by our installer

So we would not need skill catalog sources and we wouldn't have code execution of unknown origin.

For code execution we could have a plugin system, like ADK does.  They have a couple of default options: run locally, run in Vertex system, run with docker...
For OpenStack Lightspeed (and probably for OpenShift Lightspeed) we would use a  local runner or a pod runner.

And I believe having script support gives something that we don't have right now any other way: non-global tools. In this case they would be per-skill tools.

So I would vote for option B or C.

docs/design/agent-skills/agent-skills-spike.md
Akrog: +1 to path-based

docs/design/agent-skills/agent-skills-spike.md
Akrog: In ADK they multiple tools for the LLM to handle skills:
- `list_skills` 
- `load_skill`
- `load_skill_resource`
- `run_skill_script`

And they include instructions: https://github.com/google/adk-python/blob/909a8c2ad4d06ad485173f40f278c503cd66a063/src/google/adk/tools/skill_toolset.py#L60-L87

docs/design/agent-skills/agent-skills-spike.md
```diff
@@ -0,0 +1,361 @@
...
+
+This keeps the base context small while giving the LLM access to specialized knowledge on demand.
+
+### Decision 7: Configuration structure
+
+Skills are configured as a list in `lightspeed-stack.yaml`, following the `mcp_servers` pattern:
+
+```yaml
+skills:
+  - name: "code-review"
```
Akrog: -1:  I would prefer not duplicating the name and description that is already present in the `SKILL.md`.  It adds complexity to the configuration process, additional checks to the code that needs to confirm that this matches the metadata in the `SKILL.md` file...

In my opinion it should be either:

```yaml
skills:
- /var/skills/code-revie
- /var/skill/troubleshooting
```

Or if we may want in the future to have a `recursive_path` or something else:
```yaml
skills:
- path: /var/skills/code-revie
- path: /var/skill/troubleshooting
```

docs/design/agent-skills/agent-skills-spike.md
```diff
@@ -0,0 +1,361 @@
...
+Read the "Configuration" section in docs/design/agent-skills/agent-skills.md.
+Key files: src/models/config.py (around line 1817, Configuration class).
+Follow the MCP server config pattern (ModelContextProtocolServer class, line 468).
+```
+
+<!-- type: Task -->
+<!-- key: LCORE-???? -->
+### LCORE-???? Implement skill catalog injection
+
+**Description**: Inject the skill catalog (name + description for each skill) into the system prompt so the LLM knows what skills are available.
```
Akrog: Not sure if it would be better to have the `list_skills` tool instead as discussed above.

Akrog (changes requested):
Thank you for starting this work. :-)

docs/design/agent-skills/agent-skills-spike.md
umago: Agree on the terminology change here. I also initially read "customers" as the final customers and not a Lightspeed flavor (RHEL LS, OpenStack LS, etc...)

docs/design/agent-skills/agent-skills-spike.md
bostrt: Ask Red Hat (access.redhat.com/ask) injects the Skill name + description into a dedicated Skill Router's System Prompt, which is very effective for a finite number of Skills and I believe is proportional to the model's size + quality (i.e. you can't rely on 100 Skills in a small language model system prompt). I just want to add this in the phase 1 vs 2 discussion.

Something else I want to share is Ask Red Hat opted to use Structured Output to choose skills instead of Tool Calling (we tried both). We've struggled with reliable tool calling across the open-weight models we tested and Structured Output really helped a lot. We do lag behind on testing frontier models though so take with a grain of salt. 

EDIT: The Structured Output format is like: 

```
{
    "selected_skill_names": [
        "cve-search",
    ],
    "reasoning": "The user is seeking assistance with CVE-2026-.... This falls under cve-search ... blah blah "
}
```

For transparency, Ask Red Hat is currently using a Small Language Model (Granite 4.0 3B Micro) for Skill selection, which keeps routing fast and cheap but means our constraints may be more extreme than others.

bostrt: I want to clarify that Ask Red Hat uses a dedicated Skill Router agent (the SLM) that runs before the final answer generation agent — it's a two-stage LLM pipeline rather than the main agent activating skills itself via tool calls. This is a fundamentally different approach from what's being proposed here, but it's worked well for us.

bostrt:
One thing worth considering (maybe out of scope for this design?) is observability/tracing for skill activation decisions.

docs/design/agent-skills/agent-skills-spike.md
anik120: Sounds like option C is the best route here - start with no script execution, roll it out to product teams, and then iterate on script execution after MVP is in the hands of product teams and we have a chance to take in feedback.

docs/design/agent-skills/agent-skills-spike.md
anik120: From @Akrog's comment below: 

> Or if we may want in the future to have a recursive_path or something else:
>
>skills:
>- path: /var/skills/code-review
>- path: /var/skills/troubleshooting


☝🏽 looks like the best direction to move forward with - This way the API we expose doesn't have to change, and we can add `recursive_path` as an implementation detail after. 

ie you can either choose to configure as 

```
skills:
- path: /var/skills/code-review
- path: /var/skills/troubleshooting
```

or 

```
skills:
- path: /var/skills/
```

but we'll add the second capability later since that'll involve a bit of "path content validation" logic.

docs/design/agent-skills/agent-skills-spike.md
anik120: This is a lot of GREAT feedback thank you all. 

@jrobertboos looks like we can try what ADK does, and still keep phase 1 implementation scoped? I'd capture what Ask Red Hat does as a "Out of scope for phase 1 but follow up to be explored in phase 2" to account for the fact that we'll have to experiment with large language models and see the results.

tisnik (approved):
LGTM

PR #1562, opened 2026-04-21
LCORE-1957: Removed dev-tools/mcp-server

jrobertboos:
## Description

Consolidates the MCP mock server infrastructure by removing the duplicate `dev-tools/mcp-mock-server` and standardizing on the `tests/e2e/mock_mcp_server` implementation. This simplifies the test infrastructure by:

- Removing the entire `dev-tools/mcp-mock-server` directory and its associated documentation
- Updating the `mock-mcp` service to use port 3000 (previously used by the removed server)
- Removing `mcp-mock-server` service dependencies from docker-compose files
- Renaming config file `mcp-mock-server.yaml` to `mock-mcp.yaml`
- Moving the MCP client auth options test from `info.feature` to `mcp.feature` for better test organization

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [x] Documentation Update
- [x] Configuration Update
- [ ] Bump-up service version
- [x] CI configuration change
- [ ] Unit tests improvement
- [x] End to end tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Assisted-by: Claude Code
- Generated by: N/A

- Related Issue # LCORE-1957
- Closes #

## Testing

1. Run e2e tests with docker-compose to verify mock-mcp service works on port 3000:
   ```bash
   docker-compose up -d
   uv run make test-e2e
   ```
2. Verify MCP-related test scenarios pass, particularly:
  - MCP OAuth authentication tests
  - MCP client auth options endpoint test (now in mcp.feature)
  - MCP server management API tests
3. Confirm the mcp-mock-server service is no longer referenced in any configuration


jrobertboos:
**Please Review**
- @tisnik 
- @radofuchs

docker-compose-library.yaml
```diff
@@ -106,11 +87,11 @@ services:
       dockerfile: Dockerfile
     container_name: mock-mcp
     ports:
-      - "3001:3001"
+      - "3000:3000"
```
radofuchs, line 90: why did you do this change? it would be way easier to keep 3001 and not change it everywhere

tests/e2e/configuration/server-mode/lightspeed-stack.yaml
```diff
@@ -32,22 +32,4 @@ byok_rag:
 
 rag:
   tool:
-    - e2e-test-docs
```
radofuchs, line 35: why did you remove this, is it really not used anywhere?

radofuchs (changes requested):

docker-compose-library.yaml
jrobertboos: Yea but I wanted to use 3000 originally but had to switch to 3001 because the dev tools mock mcp. I know it would be easier but I think this is better.

tests/e2e/configuration/server-mode/lightspeed-stack.yaml
jrobertboos: well the e2e-test-docs were not removed (they are just at the bottom, weird diff) but yea the mcp_servers were not used at all except for testing the mcp-auth which I switched to using the lightspeed-stack-mcp.yaml instead of the default.

radofuchs: yes, the point was for the mcp part. So that actually means that you created separate lightspeed-stack config while it could have been part of the default one?

jrobertboos: well it already existed so I didn't create anything but it has to be removed from the default or else it would cause failures as auth needs to be passed to the mcp server and we don't do that in every query/streaming_query test. I also like separating out the mcp servers into a separate config as it reduces failure points.

radofuchs (approved):
LGTM

tisnik (approved):
LGTM

PR #1563, opened 2026-04-21
RSPEED-2885: sanitize model and MCP output in all response paths

Lifto:
## RSPEED-2885: sanitize model and MCP output in all response paths

QE found that the non-streaming `/v1/responses` path still leaks server internals:
- `model`: exposes full provider routing path (`google-vertex/publishers/google/models/gemini-2.5-flash`)
- `output`: includes `mcp_list_tools` items with full server-side tool definitions

**Root cause:** `_sanitize_response_dict()` only handled `instructions` and `tools`. The streaming path had ad-hoc `output` filtering at a separate call site that the non-streaming path lacked. Neither path sanitized `model`.

**Fix:**
- Move `output` array filtering into `_sanitize_response_dict()` — both paths now strip `mcp_list_tools`, `mcp_call`, `mcp_approval_request` items from server-deployed MCP servers
- Add `model` sanitization — strips provider prefix, keeps model name (e.g. `gemini-2.5-flash`)
- Remove redundant ad-hoc filtering from the streaming generator

**Tested:**
- Unit tests: 63 pass (15 for `_sanitize_response_dict` including new coverage for output and model)
- Integration tests: new tests for both streaming and non-streaming paths verifying sanitization end-to-end
- Local stack: verified with curl against both streaming and non-streaming endpoints — `model`, `instructions`, and `output` all sanitized correctly

## Summary by CodeRabbit

* **Improvements**
  * Responses now preserve client-specified model names verbatim; when the server selects a model, its provider-routing prefix is removed for a cleaner model value.
  * Centralized sanitization ensures sensitive server-deployed tool/approval items are removed from response output consistently across streaming and non-streaming responses.

* **Tests**
  * Expanded unit, integration, and end-to-end tests to validate output filtering and model echo/stripping behavior for both streaming and non-streaming paths.

* **Documentation**
  * Clarified model-in-response semantics and added examples.

### DIFF
--- a/src/app/endpoints/responses.py
+++ b/src/app/endpoints/responses.py
@@ -639,7 +639,12 @@ def _sanitize_response_dict(
       they were used as-is, the value is left unchanged.
     - ``tools``: server-deployed MCP tool definitions are removed; client-
       provided tools (those whose ``server_label`` is not in
-      ``configured_mcp_labels``) are preserved
+      ``configured_mcp_labels``) are preserved.
+    - ``output``: server-deployed MCP output items (``mcp_list_tools``,
+      ``mcp_call``, ``mcp_approval_request``) are stripped so clients only
+      see item types they understand (``message``, ``function_call``, etc.).
+    - ``model``: the provider routing prefix (everything before the last
+      ``/``) is stripped so clients see only the model name.
 
     Args:
         response_dict: Mutable dict produced by ``model_dump`` on a response
@@ -659,6 +664,17 @@ def _sanitize_response_dict(
             for tool in tools
             if tool.get("server_label") not in configured_mcp_labels
         ]
+
+    if output := response_dict.get("output"):
+        response_dict["output"] = [
+            item
+            for item in output
+            if not _is_server_mcp_output_item(item, configured_mcp_labels)
+        ]
+
+    model = response_dict.get("model")
+    if model and "/" in model:
+        response_dict["model"] = model.rsplit("/", 1)[-1]
 
 
 def _is_server_mcp_output_item(
@@ -835,16 +851,6 @@ async def response_generator(
                         configuration.rag_id_mapping,
                     )
                 )
-            # Remove server-deployed MCP items from the output array so
-            # clients only see item types they understand (message, function_call, etc.)
-            output = chunk_dict["response"].get("output")
-            if output is not None:
-                chunk_dict["response"]["output"] = [
-                    item
-                    for item in output
-                    if not _is_server_mcp_output_item(item, configured_mcp_labels)
-                ]
-
         # Intermediate response - no quota consumption and text yet
         if event_type == "response.in_progress":
             chunk_dict["response"]["available_quotas"] = {}
--- a/src/app/endpoints/responses.py
+++ b/src/app/endpoints/responses.py
@@ -282,10 +282,12 @@ async def responses_endpoint_handler(
 
     # LCORE-specific: Automatically select model if not provided in request
     # This extends the base LLS API which requires model to be specified.
+    client_model = responses_request.model
     if not responses_request.model:
         responses_request.model = await select_model_for_responses(
             client, response_context.user_conversation
         )
+    model_substituted = not client_model
     if not await check_model_configured(client, responses_request.model):
         _, model_id = extract_provider_and_model_from_model_id(responses_request.model)
         error_response = NotFoundResponse(resource="model", resource_id=model_id)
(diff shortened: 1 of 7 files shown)


tisnik (approved):
LGTM

src/app/endpoints/responses.py
```diff
@@ -660,6 +665,17 @@ def _sanitize_response_dict(
...
+    if output := response_dict.get("output"):
+        response_dict["output"] = [
+            item
+            for item in output
+            if not _is_server_mcp_output_item(item, configured_mcp_labels)
+        ]
+
+    model = response_dict.get("model")
+    if model and "/" in model:
+        response_dict["model"] = model.rsplit("/", 1)[-1]
```
radofuchs: this update requires also update in the e2e tests Responses accepts string input and Streaming esponses accepts string input

radofuchs (changes requested):
LGTM, just one update in tests needed

src/app/endpoints/responses.py
Lifto: Good catch — updated. Added `{MODEL_SHORT}` placeholder (set once in `environment.py` as `default_model_short`) and updated both `responses.feature` and `responses_streaming.feature` response assertions to use it. Request bodies still send `{PROVIDER}/{MODEL}` unchanged.

asimurka (changes requested):
Why the provider stripping from model cannot be done on client side? This introduces some implicit behavior that violates the OpenResponses standard.

Lifto:
> Why the provider stripping from model cannot be done on client side? This introduces some implicit behavior that violates the OpenResponses standard.

The google-vertex/publishers/google/models/ prefix is a Llama Stack provider routing detail that exposes our cloud infrastructure. QE flagged this as an information leak (RSPEED-2885) — any API consumer can see which cloud vendor and project path we're using.

This can't be handled client-side because the data is already leaked once it leaves the server.

QE says they are willing to flex on the model leak. I can take that change out. Shall I?

asimurka:
Is this leaked only when server configured model is used? Or does the client set model per-request in such format?
If so, then can we use the same principle as we used in `instruction`? Namely, when model is configured per-request -> echo it and if the model is configured by server (empty model in request), then we can strip the provider prefix.

Lifto:
Good idea — same pattern as `instructions`. I'll add a `model_substituted` flag: if the client specified a model, echo it unchanged. If the server defaulted it (no model in request), strip the provider prefix. That way client-specified values are never modified.

asimurka:
Yes, please do it in this way and add a short note to `docs/responses.md`. I will approve it then. Thanks for patience.

Lifto:
> Yes, please do it in this way and add a short note to `docs/responses.md`. I will approve it then. Thanks for patience.

changes are in! thank you

asimurka:
@Lifto you have some conflicts. Anyway, you don't have to update e2e tests now because the model is specified in the request so the prefix won't be stripped.

Lifto:
> @Lifto you have some conflicts. Anyway, you don't have to update e2e tests now because the model is specified in the request so the prefix won't be stripped.

The necessary changes are in, e2e tests are out. Should be ready now.

asimurka (approved):
LGTM

radofuchs:
@Lifto I would actually expect you to update the test to cover this functionality rather than removing it,. There is already a test scenario Responses omits model and auto-selects when only input is sent that should cover your new functionality

Lifto:
> @Lifto I would actually expect you to update the test to cover this functionality rather than removing it,. There is already a test scenario Responses omits model and auto-selects when only input is sent that should cover your new functionality

@radofuchs I updated the e2e tests to cover this functionality. Is this what you had in mind?

tests/unit/app/endpoints/test_responses.py
```diff
@@ -1949,26 +2020,295 @@ def test_all_fields_sanitized_together_with_substitution(self) -> None:
...
+                    },
+                    {
+                        "type": "message",
+                        "role": "assistant",
+                        "content": [{"type": "output_text", "text": "hello"}],
+                    },
+                ],
+            },
+        }
+        return completed_chunk
```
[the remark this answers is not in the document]
Lifto: Added in 3d150c9.

tests/e2e/features/responses.feature
```diff
@@ -156,6 +156,12 @@ Feature: Responses endpoint API tests
     """
     Then The status code of the response is 200
       And The body of the response contains hello
+      And the body of the response has the following structure
```
radofuchs, line 159: please align this step with the scenario Responses accepts string input above, do the same for the streaming endpoint

Lifto: Aligned with the string input scenario structure — now asserts object, status, model, and output type/role. Using `{MODEL}` (no provider prefix) for the auto-select case since the server substitutes and strips the prefix.

Lifto: @radofuchs changes applied

radofuchs (approved):
LGTM

PR #1566, opened 2026-04-21
LCORE-1723: Cross-Encoder Reranking for enhanced RAG

Anxhela21:
## Description

Cross-Encoder Reranking for Enhanced RAG Quality

  Summary

  Implemented neural reranking using cross-encoder models to improve RAG chunk relevance scoring. This enhancement replaces simple vector similarity with semantic relevance scoring when combining results from multiple knowledge sources (BYOK + OKP) that have different search types e.g. hybrid, semantic, etc

  Key Features

  - Cross-encoder model integration using sentence-transformers
  - Configurable model selection (default: cross-encoder/ms-marco-MiniLM-L6-v2)
  - Query-document pair scoring for improved semantic relevance
  - Efficient model caching to avoid reloading between requests

Flexible Configuration

  reranker:
    enabled: true
    model: "cross-encoder/ms-marco-MiniLM-L6-v2"
    top_k_multiplier: 2.0  # Fetch 2x chunks, rerank, keep top_k
    byok_boost: 1.2        # Boost factor for BYOK sources
    okp_boost: 1.0         # Boost factor for OKP sources

Auto-Enable

  - Automatically enables when both BYOK and OKP RAG sources are configured
  - Only when user hasn't explicitly configured reranker settings (see above)

 Error Handling

  - Fallback to original vector scores if reranking fails
  - Lazy model loading with error recovery
  - Model caching

  Technical Implementation

  Modified Files

  - src/utils/vector_search.py - Core reranking logic and pipeline integration
  - src/models/config.py - Configuration schema and validation
  - src/configuration.py - Configuration property access
  - src/constants.py - Default boost factors and settings
  - pyproject.toml - Added sentence-transformers dependency

  

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Related Issue #
- Closes #

## Testing

## Summary by CodeRabbit

* **New Features**
  * Optional cross-encoder RAG chunk reranking for improved relevance.
  * Reranker settings exposed in app configuration (enable/disable, model); auto-enabled when BYOK + inline RAG OKP are detected.
  * BYOK-specific rerank boost applied after reranking, with robust fallback to original ordering if the reranker is unavailable.

* **Chores**
  * Adds runtime support for cross-encoder reranking.

* **Tests**
  * New and expanded unit/integration tests covering reranking, boosting, caching, serialization, and fallbacks.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,11 @@ dependencies = [
     "jinja2>=3.1.0",
     # To be able to fix multiple CVEs, also LCORE-1117
     "requests>=2.33.0",
+    # Used for RAG chunk reranking (cross-encoder)
+    "sentence-transformers>=2.0.0",
     "datasets>=4.7.0",
+    # Used for RAG chunk reranking (cross-encoder)
+    "sentence-transformers>=2.0.0",
     # Used for error tracking and monitoring
     "sentry-sdk[fastapi]>=2.58.0",
     "python-dotenv>=1.2.2",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,10 +73,7 @@ dependencies = [
     # To be able to fix multiple CVEs, also LCORE-1117
     "requests>=2.33.0",
     # Used for RAG chunk reranking (cross-encoder)
-    "sentence-transformers>=2.0.0",
     "datasets>=4.7.0",
-    # Used for RAG chunk reranking (cross-encoder)
-    "sentence-transformers>=2.0.0",
     # Used for error tracking and monitoring
     "sentry-sdk[fastapi]>=2.58.0",
     "python-dotenv>=1.2.2",
--- a/src/models/config.py
+++ b/src/models/config.py
@@ -1817,6 +1817,63 @@ class OkpConfiguration(ConfigurationBase):
     )
 
 
+class RerankerConfiguration(ConfigurationBase):
+    """Reranker configuration for RAG chunk reranking."""
+
+    enabled: bool = True
+    model: str = Field(
+        default="cross-encoder/ms-marco-MiniLM-L6-v2",
+        title="Reranker model",
+        description="Cross-encoder model name for reranking RAG chunks. "
+        "Defaults to 'cross-encoder/ms-marco-MiniLM-L6-v2' from sentence-transformers.",
+    )
+    model_id: str = "meta-llama/llama-cross-encoder-base"
+    provider_id: str = "meta-reference"
+    top_k_multiplier: float = 2.0  # fetch 2x, rerank, keep top_k
+    byok_boost: float = 1.2
+    okp_boost: float = 1.0
+
+    # Private attribute to track if this was explicitly configured
+    _explicitly_configured: bool = PrivateAttr(default=False)
+
+    @model_validator(mode="after")
+    def mark_as_explicitly_configured(self) -> Self:
+        """Mark this configuration as explicitly set when instantiated from user input."""
+        # Only mark as explicitly configured if we're not using all default values
+        # This allows auto-enabling when user hasn't touched reranker settings
+        # Check if any field differs from default values
+        default_model = "cross-encoder/ms-marco-MiniLM-L6-v2"
+        default_model_id = "meta-llama/llama-cross-encoder-base"
+        default_provider_id = "meta-reference"
+        default_top_k_multiplier = 2.0
+        default_byok_boost = 1.2
+        default_okp_boost = 1.0
+
+        # Check if any setting differs from defaults (indicates explicit configuration)
+        current_values = [
+            self.enabled,
+            self.model,
+            self.model_id,
+            self.provider_id,
+            self.top_k_multiplier,
+            self.byok_boost,
+            self.okp_boost,
+        ]
+        default_values = [
+            True,
+            default_model,
+            default_model_id,
+            default_provider_id,
+            default_top_k_multiplier,
+            default_byok_boost,
+            default_okp_boost,
+        ]
+
+        if current_values != default_values:
+            self._explicitly_configured = True
+        return self
+
+
 class AzureEntraIdConfiguration(ConfigurationBase):
     """Microsoft Entra ID authentication attributes for Azure."""
 
@@ -1976,6 +2033,12 @@ class Configuration(ConfigurationBase):
         "in rag.inline or rag.tool.",
     )
 
+    reranker: RerankerConfiguration = Field(
+        default_factory=RerankerConfiguration,
+        title="Reranker configuration",
+        description="Configuration for neural reranking of RAG chunks using cross-encoder.",
+    )
+
     @model_validator(mode="after")
     def validate_mcp_auth_headers(self) -> Self:
         """
@@ -2075,6 +2138,48 @@ class Configuration(ConfigurationBase):
                 "are configured.",
                 quota_subject,
             )
+
+        return self
+
+    @model_validator(mode="after")
+    def validate_reranker_auto_enable(self) -> Self:
+        """Automatically enable reranker when both BYOK and OKP RAG are configured.
+
+        When users have both BYOK (Bring Your Own Key) entries in byok_rag and OKP
+        (OpenShift Knowledge Platform) configured in the RAG strategies, automatically
+        enable the reranker if it's not explicitly disabled. This improves result
+        quality when multiple knowledge sources are available.
+
+        Returns:
+            Self: The validated configuration instance with reranker potentially enabled.
+        """
+        # Check if BYOK RAG entries are configured
+        has_byok = len(self.byok_rag) > 0
+
+        # Check if OKP is configured in either inline or tool RAG strategies
+        # pylint: disable=no-member
+        has_okp = (
+            constants.OKP_RAG_ID in self.rag.inline
+            or constants.OKP_RAG_ID in self.rag.tool
+        )
+
+        # If both BYOK and OKP are present and reranker is using default settings,
+        # ensure it's enabled for optimal results
+        if (
+            has_byok
+            and has_okp
+            and not hasattr(self.reranker, "_explicitly_configured")
+        ):
+            # pylint: disable=no-member
+            if not self.reranker.enabled:
+                logger.info(
+                    "Automatically enabling reranker: Both BYOK RAG (%d entries) and OKP "
+                    "are configured. Reranking improves result quality when multiple "
+                    "knowledge sources are available.",
+                    len(self.byok_rag),
+                )
+                # pylint: disable=no-member
+                self.reranker.enabled = True
 
         return self
 
--- a/src/models/config.py
+++ b/src/models/config.py
@@ -1827,8 +1827,6 @@ class RerankerConfiguration(ConfigurationBase):
         description="Cross-encoder model name for reranking RAG chunks. "
         "Defaults to 'cross-encoder/ms-marco-MiniLM-L6-v2' from sentence-transformers.",
     )
-    model_id: str = "meta-llama/llama-cross-encoder-base"
-    provider_id: str = "meta-reference"
     top_k_multiplier: float = 2.0  # fetch 2x, rerank, keep top_k
     byok_boost: float = 1.2
     okp_boost: float = 1.0
@@ -1843,8 +1841,6 @@ class RerankerConfiguration(ConfigurationBase):
         # This allows auto-enabling when user hasn't touched reranker settings
         # Check if any field differs from default values
         default_model = "cross-encoder/ms-marco-MiniLM-L6-v2"
-        default_model_id = "meta-llama/llama-cross-encoder-base"
-        default_provider_id = "meta-reference"
         default_top_k_multiplier = 2.0
         default_byok_boost = 1.2
         default_okp_boost = 1.0
@@ -1853,8 +1849,6 @@ class RerankerConfiguration(ConfigurationBase):
         current_values = [
             self.enabled,
             self.model,
-            self.model_id,
-            self.provider_id,
             self.top_k_multiplier,
             self.byok_boost,
             self.okp_boost,
@@ -1862,8 +1856,6 @@ class RerankerConfiguration(ConfigurationBase):
         default_values = [
             True,
             default_model,
-            default_model_id,
-            default_provider_id,
             default_top_k_multiplier,
             default_byok_boost,
             default_okp_boost,
--- a/src/models/config.py
+++ b/src/models/config.py
@@ -1820,16 +1820,18 @@ class OkpConfiguration(ConfigurationBase):
 class RerankerConfiguration(ConfigurationBase):
     """Reranker configuration for RAG chunk reranking."""
 
-    enabled: bool = True
+    enabled: bool = Field(
+        default=False,
+        title="Reranker enabled",
+        description="When True, reranking applied to RAG chunks. "
+        "When False, reranking is disabled and original scoring used.",
+    )
     model: str = Field(
         default="cross-encoder/ms-marco-MiniLM-L6-v2",
         title="Reranker model",
         description="Cross-encoder model name for reranking RAG chunks. "
         "Defaults to 'cross-encoder/ms-marco-MiniLM-L6-v2' from sentence-transformers.",
     )
-    top_k_multiplier: float = 2.0  # fetch 2x, rerank, keep top_k
-    byok_boost: float = 1.2
-    okp_boost: float = 1.0
 
     # Private attribute to track if this was explicitly configured
     _explicitly_configured: bool = PrivateAttr(default=False)
@@ -1837,32 +1839,9 @@ class RerankerConfiguration(ConfigurationBase):
     @model_validator(mode="after")
     def mark_as_explicitly_configured(self) -> Self:
         """Mark this configuration as explicitly set when instantiated from user input."""
-        # Only mark as explicitly configured if we're not using all default values
-        # This allows auto-enabling when user hasn't touched reranker settings
-        # Check if any field differs from default values
-        default_model = "cross-encoder/ms-marco-MiniLM-L6-v2"
-        default_top_k_multiplier = 2.0
-        default_byok_boost = 1.2
-        default_okp_boost = 1.0
-
-        # Check if any setting differs from defaults (indicates explicit configuration)
-        current_values = [
-            self.enabled,
-            self.model,
-            self.top_k_multiplier,
-            self.byok_boost,
-            self.okp_boost,
-        ]
-        default_values = [
-            True,
-            default_model,
-            default_top_k_multiplier,
-            default_byok_boost,
-            default_okp_boost,
-        ]
-
-        if current_values != default_values:
+        if self.model_fields_set:
             self._explicitly_configured = True
+
         return self
 
 
@@ -2137,8 +2116,8 @@ class Configuration(ConfigurationBase):
     def validate_reranker_auto_enable(self) -> Self:
         """Automatically enable reranker when both BYOK and OKP RAG are configured.
 
-        When users have both BYOK (Bring Your Own Key) entries in byok_rag and OKP
-        (OpenShift Knowledge Platform) configured in the RAG strategies, automatically
+        When users have both BYOK entries in byok_rag and OKP
+        configured in the RAG strategies, automatically
         enable the reranker if it's not explicitly disabled. This improves result
         quality when multiple knowledge sources are available.
 
@@ -2150,28 +2129,23 @@ class Configuration(ConfigurationBase):
 
         # Check if OKP is configured in either inline or tool RAG strategies
         # pylint: disable=no-member
-        has_okp = (
-            constants.OKP_RAG_ID in self.rag.inline
-            or constants.OKP_RAG_ID in self.rag.tool
-        )
+        has_okp = constants.OKP_RAG_ID in self.rag.inline
 
         # If both BYOK and OKP are present and reranker is using default settings,
         # ensure it's enabled for optimal results
         if (
             has_byok
             and has_okp
-            and not hasattr(self.reranker, "_explicitly_configured")
+            and not self.reranker._explicitly_configured  # pylint: disable=protected-access
+            and not self.reranker.enabled
         ):
-            # pylint: disable=no-member
-            if not self.reranker.enabled:
-                logger.info(
-                    "Automatically enabling reranker: Both BYOK RAG (%d entries) and OKP "
-                    "are configured. Reranking improves result quality when multiple "
-                    "knowledge sources are available.",
-                    len(self.byok_rag),
-                )
-                # pylint: disable=no-member
-                self.reranker.enabled = True
+            logger.info(
+                "Automatically enabling reranker: Both BYOK RAG (%d entries) or "
+                "other inline RAG and OKP are configured. Reranking improves result "
+                "quality when multiple knowledge sources are available.",
+                len(self.byok_rag),
+            )
+            self.reranker.enabled = True
 
         return self
 
--- a/src/utils/vector_search.py
+++ b/src/utils/vector_search.py
@@ -21,188 +21,52 @@ from log import get_logger
 from models.common.query import SolrVectorSearchRequest
 from models.common.responses.types import ResponseInput
 from models.common.turn_summary import RAGChunk, RAGContext, ReferencedDocument
+from utils.reranker import apply_byok_rerank_boost, rerank_chunks_with_cross_encoder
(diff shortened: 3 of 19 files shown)


src/models/config.py
```diff
@@ -1809,6 +1809,63 @@ class OkpConfiguration(ConfigurationBase):
...
+        default="cross-encoder/ms-marco-MiniLM-L6-v2",
+        title="Reranker model",
+        description="Cross-encoder model name for reranking RAG chunks. "
+        "Defaults to 'cross-encoder/ms-marco-MiniLM-L6-v2' from sentence-transformers.",
+    )
+    model_id: str = "meta-llama/llama-cross-encoder-base"
+    provider_id: str = "meta-reference"
+    top_k_multiplier: float = 2.0  # fetch 2x, rerank, keep top_k
+    byok_boost: float = 1.2
+    okp_boost: float = 1.0
```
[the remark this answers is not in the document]
Anxhela21: removed.

src/models/config.py
```diff
@@ -1809,6 +1809,63 @@ class OkpConfiguration(ConfigurationBase):
...
+            default_model_id,
+            default_provider_id,
+            default_top_k_multiplier,
+            default_byok_boost,
+            default_okp_boost,
+        ]
+
+        if current_values != default_values:
+            self._explicitly_configured = True
+        return self
```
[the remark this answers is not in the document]
Anxhela21: Good point. Done.

src/utils/vector_search.py
```diff
@@ -25,6 +25,185 @@
...
+            )
+
+            _cross_encoder_models[model_name] = CrossEncoder(model_name)
+            logger.info("Loaded cross-encoder for RAG reranking: %s", model_name)
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            logger.warning(
+                "Could not load cross-encoder for reranking (%s): %s", model_name, e
+            )
+            _cross_encoder_models[model_name] = None
+    return _cross_encoder_models[model_name]
```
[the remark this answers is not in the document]
are-ces: @syedriko or @tisnik do you think this is a valid concern?

src/utils/vector_search.py
```diff
@@ -25,6 +25,185 @@
...
+        logger.warning(
+            "Cross-encoder reranking failed, falling back to original scoring: %s", e
+        )
+        # Fallback: sort by original score and take top_k
+        sorted_chunks = sorted(
+            chunks,
+            key=lambda c: c.score if c.score is not None else float("-inf"),
+            reverse=True,
+        )
+        return sorted_chunks[:top_k]
```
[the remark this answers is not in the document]
are-ces: This seems a valid concern

src/utils/vector_search.py
```diff
@@ -25,6 +25,185 @@
...
+        # Get the cached cross-encoder model
+        model = _get_cross_encoder(model_name)
+        if model is None:
+            raise RuntimeError(f"Failed to load cross-encoder model: {model_name}")
+
+        logger.debug("Using cross-encoder model: %s", model_name)
+
+        # Create query-chunk pairs for scoring
+        pairs = [(query, chunk.content) for chunk in chunks]
+        scores = model.predict(pairs)
```
are-ces: As CodeRabbit points out, this is synchronous and blocks the process, please make async

src/utils/vector_search.py
```diff
@@ -527,51 +719,70 @@ async def _fetch_solr_rag(
...
-    # Fetch from all enabled RAG sources in parallel
-    byok_chunks_task = _fetch_byok_rag(client, query, vector_store_ids)
+    # Fetch 2*BYOK_RAG_MAX_CHUNKS from each source in parallel
+    byok_chunks_task = _fetch_byok_rag(
+        client, query, vector_store_ids, max_chunks=pool_size
+    )
     solr_chunks_task = _fetch_solr_rag(client, query, solr)
 
-    (byok_chunks, byok_docs), (solr_chunks, solr_docs) = await asyncio.gather(
+    (byok_chunks, _), (solr_chunks, _) = await asyncio.gather(
```
are-ces: The new `_referenced_documents_from_rag_chunks` (line 169) duplicates the same extraction logic already present in `_process_byok_rag_chunks_for_documents` and `_process_solr_chunks_for_documents`.

On top of that, those two per-source functions are effectively dead code in the current flow, their return values are discarded on line 754:

```python
(byok_chunks, _), (solr_chunks, _) = await asyncio.gather(
    byok_chunks_task, solr_chunks_task
)
```

...and then `_referenced_documents_from_rag_chunks` re-extracts the same documents from the final chunks at line 785.

Suggestion: Instead of discarding the per-source documents and re-extracting them, keep the documents returned by `_fetch_byok_rag` and `_fetch_solr_rag`, then after reranking, filter/reorder them to match whichever chunks survived the rerank. This avoids duplicating the extraction logic and is more efficient — one lookup pass instead of a second extraction. `_referenced_documents_from_rag_chunks` can then be removed.

src/models/config.py
```diff
@@ -2072,6 +2101,46 @@ def validate_rlsapi_v1_quota_configuration(self) -> Self:
...
+            Self: The validated configuration instance with reranker potentially enabled.
+        """
+        # Check if BYOK RAG entries are configured
+        has_byok = len(self.byok_rag) > 0
+
+        # Check if OKP is configured in either inline or tool RAG strategies
+        # pylint: disable=no-member
+        has_okp = (
+            constants.OKP_RAG_ID in self.rag.inline
+            or constants.OKP_RAG_ID in self.rag.tool
```
are-ces: Why should the re-ranked be active if OKP is in RAG as a Tool?

src/models/config.py
```diff
@@ -2072,6 +2101,46 @@ def validate_rlsapi_v1_quota_configuration(self) -> Self:
...
+        When users have both BYOK (Bring Your Own Key) entries in byok_rag and OKP
+        (OpenShift Knowledge Platform) configured in the RAG strategies, automatically
+        enable the reranker if it's not explicitly disabled. This improves result
+        quality when multiple knowledge sources are available.
+
+        Returns:
+            Self: The validated configuration instance with reranker potentially enabled.
+        """
+        # Check if BYOK RAG entries are configured
+        has_byok = len(self.byok_rag) > 0
```
are-ces, line 2128: Should also check that there are any entries registered in `inline RAG`

src/utils/vector_search.py
```diff
@@ -527,51 +719,70 @@ async def _fetch_solr_rag(
...
         client: The AsyncLlamaStackClient to use for the request
-        moderation_decision: The moderation decision
         query: The user's query
         vector_store_ids: The vector store IDs to query
         solr: Structured Solr inline RAG request from the API (optional).
 
     Returns:
         RAGContext containing formatted context text and referenced documents
     """
-    if moderation_decision == "blocked":
```
are-ces, line 551: Why was this removed? This if was there to prevent doing RAG in case of an unsafe input

src/utils/vector_search.py
```diff
@@ -451,17 +639,20 @@ async def _fetch_byok_rag(
     return rag_chunks, referenced_documents
 
 
-async def _fetch_solr_rag(
+async def _fetch_solr_rag(  # pylint: disable=too-many-locals
     client: AsyncLlamaStackClient,
     query: str,
     solr: Optional[SolrVectorSearchRequest] = None,
+    max_chunks: Optional[int] = None,
```
are-ces: This parameter is not used

src/utils/vector_search.py
```diff
@@ -527,51 +719,70 @@ async def _fetch_solr_rag(
...
+    pool_size = 2 * constants.BYOK_RAG_MAX_CHUNKS
+    top_k = constants.BYOK_RAG_MAX_CHUNKS
 
-    # Fetch from all enabled RAG sources in parallel
-    byok_chunks_task = _fetch_byok_rag(client, query, vector_store_ids)
+    # Fetch 2*BYOK_RAG_MAX_CHUNKS from each source in parallel
+    byok_chunks_task = _fetch_byok_rag(
+        client, query, vector_store_ids, max_chunks=pool_size
+    )
     solr_chunks_task = _fetch_solr_rag(client, query, solr)
```
are-ces, line 636: Why are we not increasing the pool size for OKP chunks? I might miss something here...

src/utils/vector_search.py
```diff
@@ -25,6 +25,185 @@
...
+
+    Args:
+        model_name: Name of the cross-encoder model to load.
+
+    Returns:
+        Loaded CrossEncoder model instance, or None if loading fails.
+    """
+    if model_name not in _cross_encoder_models:
+        try:
+            from sentence_transformers import (  # pylint: disable=import-outside-toplevel
```
are-ces: This dep should be moved on top

are-ces (changes requested):
The example config in the PR description must be outdated because
```
top_k_multiplier: 2.0 # Fetch 2x chunks, rerank, keep top_k
byok_boost: 1.2 # Boost factor for BYOK sources
okp_boost: 1.0 # Boost factor for OKP sources
```
are missing from the configuration in the PR. 

Would be nice to see a configured reranker in the example configuration `examples/lightspeed-stack-byok-okp-rag.yaml`.

Please update unit and integration tests; fix Black and Linting failing tests. TY!

pyproject.toml
```diff
@@ -72,6 +72,8 @@ dependencies = [
     "jinja2>=3.1.0",
     # To be able to fix multiple CVEs, also LCORE-1117
     "requests>=2.33.0",
+    # Used for RAG chunk reranking (cross-encoder)
+    "sentence-transformers>=2.0.0",
```
are-ces: Should the `sentence-transformer` dependency be added @syedriko @tisnik? In that case uv.lock will need to be updated too

src/models/config.py
```diff
@@ -1812,6 +1812,29 @@ class OkpConfiguration(ConfigurationBase):
     )
 
 
+class RerankerConfiguration(ConfigurationBase):
+    """Reranker configuration for RAG chunk reranking."""
+
+    enabled: bool = True
```
tisnik: nit: can be initialized as `Field` so it would be possible to specify title and description.

pyproject.toml
tisnik: yes, it needs to be updated to pass CI etc.

src/models/config.py
Anxhela21: Oh good point!

src/models/config.py
Anxhela21: this is looking if okp is enabled either as inline or tool since we can configure okp either way unless I'm mistaken?

pyproject.toml
Anxhela21: thanks missed that - will update uv.lock

src/models/config.py
Anxhela21: I suppose for simplicity we can avoid checking whether it's active as tool.

Anxhela21: removed for now.

src/models/config.py
Anxhela21: done.

src/models/config.py
Anxhela21: done.

pyproject.toml
Anxhela21: done.

src/utils/vector_search.py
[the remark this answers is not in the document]
Anxhela21: Done.

src/utils/vector_search.py
Anxhela21: Done.

src/utils/vector_search.py
Anxhela21: this was unintentional. restored. thanks for catching!

src/utils/vector_search.py
Anxhela21: because we only want to compound the weight for BYOK chunks so we prioritize them.

src/utils/vector_search.py
Anxhela21: removed.

src/utils/vector_search.py
Anxhela21: Done.

src/models/config.py
```diff
@@ -2072,6 +2105,49 @@ def validate_rlsapi_v1_quota_configuration(self) -> Self:
...
+
+        Returns:
+            Self: The validated configuration instance with reranker potentially enabled.
+        """
+        # Check if BYOK RAG entries are configured
+        has_byok = len(self.byok_rag) > 0
+
+
+        #Check if inline RAG
+        has_inline_rag = len(self.rag) > 0
```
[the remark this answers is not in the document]
Anxhela21: Done.

src/utils/vector_search.py
```diff
@@ -527,51 +756,80 @@ async def _fetch_solr_rag(
...
+        )
+        context_chunks = _apply_byok_rerank_boost(reranked)[:top_k]
+        logger.info(
+            "Reranker completed: returned %d top chunks after BYOK boost",
+            len(context_chunks)
+        )
+    else:
+        logger.info("Reranker disabled: using original vector similarity scores")
+        # Skip reranking, just apply boost and take top_k from original scores
+        context_chunks = _apply_byok_rerank_boost(merged)[:top_k]
```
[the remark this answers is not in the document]
Anxhela21: Done.

src/utils/vector_search.py
are-ces: Right, sounds good!

src/models/config.py
```diff
@@ -1812,6 +1812,33 @@ class OkpConfiguration(ConfigurationBase):
...
 
 
+class RerankerConfiguration(ConfigurationBase):
+    """Reranker configuration for RAG chunk reranking."""
+
+    enabled: bool = Field(
+        default = True,
+        title="Reranker enabled",
+        description= "When True, reranking applied to RAG chunks. "
+        "When False, reranking is disabled and origianl scoring used.",)
```
syedriko: "When False, reranking is disabled and original scoring is used."

src/models/config.py
```diff
@@ -2072,6 +2105,49 @@ def validate_rlsapi_v1_quota_configuration(self) -> Self:
 
         return self
 
+    @model_validator(mode="after")
+    def validate_reranker_auto_enable(self) -> Self:
+        """Automatically enable reranker when both BYOK and OKP RAG are configured.
+
+        When users have both BYOK (Bring Your Own Key) entries in byok_rag and OKP
```
syedriko: Bring Your Own Knowledge

Anxhela21: good catch! 🤣

src/utils/vector_search.py
Anxhela21: Good catch, will keep those documents  and just filter the re-ranked docs to avoid duplication.

Anxhela21: Done.

src/utils/vector_search.py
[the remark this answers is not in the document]
syedriko: @are-ces Yep, absolutely real, I had to fix this very issue in ~~Llama Stack~~ OGX. @Anxhela21, please take a look at https://github.com/ogx-ai/ogx/pull/4636 and the issue that it fixes https://github.com/ogx-ai/ogx/issues/4635.

src/utils/vector_search.py
```diff
@@ -25,6 +26,234 @@
...
+            _cross_encoder_models[model_name] = None
+    return _cross_encoder_models[model_name]
+
+
+# pylint: disable=too-many-locals,too-many-branches
+async def _rerank_chunks_with_cross_encoder(
+    query: str,
+    chunks: list[RAGChunk],
+    top_k: int,
+    model_name: str = "cross-encoder/ms-marco-MiniLM-L6-v2",
```
syedriko: I'd vote for the config to be the only place that defines the default model name.

syedriko: What would be even better is to define a constant for it src/constants.py, like how it was done for the default embedding model: https://github.com/lightspeed-core/lightspeed-stack/blob/be92b891360ed07b9a81f7c4a2dbc8e266db90ab/src/constants.py#L179

src/models/config.py
```diff
@@ -1817,6 +1817,34 @@ class OkpConfiguration(ConfigurationBase):
     )
 
 
+class RerankerConfiguration(ConfigurationBase):
+    """Reranker configuration for RAG chunk reranking."""
+
+    enabled: bool = Field(
+        default=True,
```
syedriko: I don't think we should enable the reranker by default, or we run the risk of adding latency into the invocation path, unknown to the user, for no good reason.

tests/unit/models/config/test_reranker_configuration.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class TestRerankerConfiguration:
+    """Tests for RerankerConfiguration model."""
+
+    def test_default_values(self) -> None:
+        """Test that RerankerConfiguration has correct default values."""
+        config = RerankerConfiguration()
+        assert config.enabled is True
+        assert config.model == "cross-encoder/ms-marco-MiniLM-L6-v2"
```
syedriko: We can use the future constant from src/constants.py here.

tests/unit/models/config/test_reranker_configuration.py
```diff
@@ -0,0 +1,52 @@
...
+        """Test configuration with custom cross-encoder model."""
+        config = RerankerConfiguration(model="cross-encoder/ms-marco-TinyBERT-L2-v2")
+        assert config.model == "cross-encoder/ms-marco-TinyBERT-L2-v2"
+        assert config.enabled is True
+
+    def test_disabled_reranker(self) -> None:
+        """Test configuration with reranker disabled."""
+        config = RerankerConfiguration(enabled=False)
+        assert config.enabled is False
+        assert config.model == "cross-encoder/ms-marco-MiniLM-L6-v2"
```
syedriko: And here.

tests/unit/models/config/test_reranker_configuration.py
```diff
@@ -0,0 +1,52 @@
...
+    def test_default_values(self) -> None:
+        """Test that RerankerConfiguration has correct default values."""
+        config = RerankerConfiguration()
+        assert config.enabled is True
+        assert config.model == "cross-encoder/ms-marco-MiniLM-L6-v2"
+
+    def test_custom_model(self) -> None:
+        """Test configuration with custom cross-encoder model."""
+        config = RerankerConfiguration(model="cross-encoder/ms-marco-TinyBERT-L2-v2")
+        assert config.model == "cross-encoder/ms-marco-TinyBERT-L2-v2"
```
syedriko, line 22: It'd be better to define the constant model name once.

src/utils/vector_search.py
```diff
@@ -25,6 +26,234 @@
 
 logger = get_logger(__name__)
 
+# Lazy-loaded cross-encoder models for reranking RAG chunks (CPU-bound, use in thread).
```
syedriko: I suggest moving all the newly addded reranker code into a separate source file. vector_search.py is growing too large.

src/utils/vector_search.py
```diff
@@ -551,27 +793,57 @@ async def build_rag_context(
...
     )
 
-    # Merge chunks from all sources (BYOK + Solr)
-    context_chunks = byok_chunks + solr_chunks
+    # Merge: combine and sort by score, keep top 2*BYOK_RAG_MAX_CHUNKS
+    merged = byok_chunks + solr_chunks
+    merged.sort(
+        key=lambda c: c.score if c.score is not None else float("-inf"),
+        reverse=True,
+    )
```
syedriko, line 646: How can we sort by score before reranking? These scores are from different scales.

src/utils/vector_search.py
```diff
@@ -25,6 +26,232 @@
...
+    if model_name not in _cross_encoder_models:
+        try:
+            _cross_encoder_models[model_name] = CrossEncoder(model_name)
+            logger.info("Loaded cross-encoder for RAG reranking: %s", model_name)
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            logger.warning(
+                "Could not load cross-encoder for reranking (%s): %s", model_name, e
+            )
+            _cross_encoder_models[model_name] = None
+    return _cross_encoder_models[model_name]
```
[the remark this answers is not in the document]
syedriko: @Anxhela21 PTAL @ https://github.com/ogx-ai/ogx/pull/4636, the original code there offloads model loading to a thread.

src/utils/vector_search.py
Anxhela21: This is a good point. I don't think this sort is actually preserving anything useful and can be negated.

src/utils/vector_search.py
are-ces: +1 on the constant

src/utils/vector_search.py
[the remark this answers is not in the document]
Anxhela21: Done.

src/utils/vector_search.py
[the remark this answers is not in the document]
Anxhela21: Done.

src/models/config.py
```diff
@@ -2078,6 +2112,43 @@ def validate_rlsapi_v1_quota_configuration(self) -> Self:
...
+        # ensure it's enabled for optimal results
+        if has_byok or has_inline_rag and has_okp and not self.reranker.enabled:
+
+            logger.info(
+                "Automatically enabling reranker: Both BYOK RAG (%d entries) or "
+                "other inline RAG and OKP are configured. Reranking improves result "
+                "quality when multiple knowledge sources are available.",
+                len(self.byok_rag),
+            )
+            self.reranker.enabled = True
```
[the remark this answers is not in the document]
Anxhela21: Done.

src/utils/vector_search.py
Anxhela21: Done.

src/models/config.py
Anxhela21: Okay, we can set default to False for now. I will add in the docs how to enable.

src/utils/vector_search.py
Anxhela21: Good point.

are-ces:
@Anxhela21 please fix unit tests. Not sure what causes the error in the e2e tests, I suspect the failure is not caught by the output debug logs

are-ces:
In [e2e tests server mode group 3](https://github.com/lightspeed-core/lightspeed-stack/actions/runs/25461109280/job/74703214052?pr=1566) this test is failing:
`Failing scenarios:
  tests/e2e/features/inline_rag.feature:36  Inline RAG query includes referenced documents`
  
 PTAL, your changes might have to do with the failure.

pyproject.toml
are-ces: I think this line is redundant and not needed since it is already defined in `llslibdev` below.

src/utils/vector_search.py
```diff
@@ -14,6 +14,7 @@
 )
 from llama_stack_client import AsyncLlamaStackClient
 from pydantic import AnyUrl
+from sentence_transformers import CrossEncoder
```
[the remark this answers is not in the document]
are-ces: Exactly, this is why all e2e tests are failing. I tried moving this to `_get_cross_encoder` in my fork and all e2e tests start, please fix this and should fix most issues @Anxhela21.
PTAL: https://github.com/are-ces/lightspeed-stack/actions/runs/25485221307/job/74778969059

are-ces (changes requested):
PTAL I (or better Coderabbit :) might have found the issue for the e2e test failures

src/utils/vector_search.py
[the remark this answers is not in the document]
Anxhela21: I don't believe this is the fix. I still see tests failing due to referenced document

src/utils/vector_search.py
Anxhela21: actually if we remove this we loose all the okp chunks because we truncate the list before sorting by score. So we sort to prevent loosing okp chunks in the final merged list and then we rerank that list which has chunks from both sources.

src/utils/vector_search.py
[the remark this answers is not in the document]
are-ces: This fixes the e2e test failures not the reference doc! The failure you mention is caused by logic inside, I will take a look asap

are-ces (approved):

tisnik (approved):
LGTM, nice work

PR #1627, opened 2026-04-29
LCORE-2034: Unified mcp registration endpoints

asimurka:
## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [x] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Assisted-by: Cursor

- Related Issue # [LCORE-2034](https://redhat.atlassian.net/browse/LCORE-2034)
- Closes #

## Testing

## Summary by CodeRabbit

* **API Changes**
  * MCP server delete endpoint now returns HTTP 200 with structured response (`deleted` boolean and `response` message) instead of HTTP 404 for non-existent servers.
  * Statically configured MCP servers cannot be deleted and return HTTP 403.
  * Improved error details for MCP server registration failures.

### DIFF
--- a/src/app/endpoints/mcp_servers.py
+++ b/src/app/endpoints/mcp_servers.py
@@ -3,7 +3,8 @@
 from typing import Annotated, Any
 
 from fastapi import APIRouter, Depends, HTTPException, Request, status
-from llama_stack_client import APIConnectionError
+from llama_stack_api.common.errors import ToolGroupNotFoundError
+from llama_stack_client import APIConnectionError, APIStatusError
 
 from authentication import get_auth_dependency
 from authentication.interface import AuthTuple
@@ -22,7 +23,6 @@ from models.responses import (
     MCPServerInfo,
     MCPServerListResponse,
     MCPServerRegistrationResponse,
-    NotFoundResponse,
     ServiceUnavailableResponse,
     UnauthorizedResponse,
 )
@@ -37,7 +37,9 @@ register_responses: dict[int | str, dict[str, Any]] = {
     401: UnauthorizedResponse.openapi_response(examples=UNAUTHORIZED_OPENAPI_EXAMPLES),
     403: ForbiddenResponse.openapi_response(examples=["endpoint"]),
     409: ConflictResponse.openapi_response(examples=["mcp server"]),
-    500: InternalServerErrorResponse.openapi_response(examples=["configuration"]),
+    500: InternalServerErrorResponse.openapi_response(
+        examples=["configuration", "mcp server registration"]
+    ),
     503: ServiceUnavailableResponse.openapi_response(
         examples=["llama stack", "kubernetes api"]
     ),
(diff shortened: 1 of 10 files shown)


src/app/endpoints/mcp_servers.py
```diff
@@ -216,46 +213,30 @@ async def delete_mcp_server_handler(
...
+    except APIStatusError:
+        logger.warning("MCP server not found, treating as already deleted.")
 
-    logger.info("Dynamically unregistered MCP server: %s", name)
-
-    return MCPServerDeleteResponse(
-        name=name,
-        message=f"MCP server '{name}' unregistered successfully",
-    )
+    return MCPServerDeleteResponse(deleted=local_deleted, name=name)
```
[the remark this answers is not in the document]
asimurka: MCP servers are either dynamic or static. Even if not, such case would still end up returning deleted=false response.

src/app/endpoints/mcp_servers.py
```diff
@@ -216,46 +214,30 @@ async def delete_mcp_server_handler(
...
-        else:
-            response = NotFoundResponse(resource="MCP server", resource_id=name)
-        raise HTTPException(**response.model_dump())
+        static_mcp_names = {s.name for s in configuration.mcp_servers}
+        if name in static_mcp_names:
+            response = ForbiddenResponse.mcp_server_static_config(name)
+            raise HTTPException(**response.model_dump())
+
+    try:
+        configuration.remove_mcp_server(name)
```
maysunfaisal: The local removal now happens before the Llama Stack unregister call. If APIConnectionError is raised from Llama Stack, the server is removed locally but the toolgroup persists in Llama Stack, creating an inconsistent state. I would consider keeping the remote call first and only removing locally on success or confirmed not found.

src/app/endpoints/mcp_servers.py
```diff
@@ -216,46 +214,30 @@ async def delete_mcp_server_handler(
...
-            e,
-        )
-
-    try:
-        configuration.remove_mcp_server(name)
-    except ValueError as e:
-        logger.error("Failed to remove MCP server from configuration: %s", e)
-        response = NotFoundResponse(resource="MCP server", resource_id=name)
-        raise HTTPException(**response.model_dump()) from e
+    except (APIStatusError, ToolGroupNotFoundError):
```
maysunfaisal: APIStatusError covers all non-2xx responses (401, 403, 500, etc.), not just 404. A real backend failure (e.g. Llama Stack returning 500 or 403) would be silently treated as "already deleted." 

Probably lets catch ToolGroupNotFoundError, or check e.status_code == 404 within the APIStatusError handler before treating it as idempotent.

src/models/responses.py
```diff
@@ -326,19 +326,35 @@ class MCPServerListResponse(AbstractSuccessfulResponse):
     }
 
 
-class MCPServerDeleteResponse(AbstractSuccessfulResponse):
-    """Response for a successful MCP server deletion."""
+class MCPServerDeleteResponse(AbstractDeleteResponse):
```
maysunfaisal: By inheriting AbstractDeleteResponse, this class now exposes a success field that always serializes as true (see the base class success property). Combined with deleted: false for the not found case, the response becomes {"deleted": false, "success": true} which is contradictory.

Probably override success to mirror deleted?

asimurka: Created a separate PR for this, which will move the deprecated `success` flag only to dedicated model so it is not inherited: https://github.com/lightspeed-core/lightspeed-stack/pull/1672

src/app/endpoints/mcp_servers.py
asimurka: `ToolGroupNotFoundError` is raised in library mode and `NotFoundError` in server mode. That's why we keep catching both of them

tisnik (approved):
LGTM

PR #1636, opened 2026-04-29
RSPEED-2959: Add default model availability check to readiness probe

major:
## Description

The `/readiness` endpoint only verified provider health via `client.providers.list()` but did not check that the configured default model was actually registered in the Llama Stack model registry. This allowed pods where model registration failed during startup to pass readiness checks and receive traffic, returning 404 "Model not found" on every inference request.

Observed in rls-prod: one pod (hk84t) out of five passed readiness (VertexAI provider reported healthy) but returned 404 on every `/v1/infer` and `/v1/responses` request because the default model `google-vertex/publishers/google/models/gemini-2.5-flash` was missing from the in-memory model registry. The other four pods served requests normally.

This PR adds `check_default_model_available()` to `health.py` that verifies the configured default model exists in `client.models.list()`. When the model is missing, `/readiness` returns HTTP 503 so Kubernetes removes the pod from the service load balancer.

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [x] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Assisted-by: Claude (via OpenCode)
- Generated by: N/A

- Related Issue #RSPEED-2959

## Testing

1. Unit tests cover all branches of `check_default_model_available()`:
   - Model found in registry (returns ready)
   - Model missing from registry (returns 503)
   - No inference configuration (passes through)
   - No default model configured (passes through)
   - `APIConnectionError` during model list (returns 503)
   - `APIStatusError` during model list (returns 503)
2. Integration tests for the updated `readiness_probe_get_method`:
   - Providers healthy + model missing = 503
   - Providers healthy + model available = 200
3. All 15 tests in `test_health.py` pass (8 new, 7 existing unchanged)
4. `ruff check` and `ruff format` pass clean

## Summary by CodeRabbit

* **New Features**
  * Readiness now verifies the configured default model exists in the model registry; if unavailable, readiness returns HTTP 503 with a clear reason and provider status.

* **Infrastructure**
  * Client-side model-availability checks added to query the model registry and attempt recovery when possible.

* **Tests**
  * Expanded unit and integration tests covering model-availability checks, error paths, and recovery scenarios.

### DIFF
--- a/src/app/endpoints/health.py
+++ b/src/app/endpoints/health.py
@@ -15,6 +15,7 @@ from authentication import get_auth_dependency
 from authentication.interface import AuthTuple
 from authorization.middleware import authorize
 from client import AsyncLlamaStackClientHolder
+from configuration import configuration
 from log import get_logger
 from models.config import Action
 from models.responses import (
@@ -95,6 +96,31 @@ async def get_providers_health_statuses() -> list[ProviderHealthStatus]:
         ]
 
 
+async def check_default_model_available() -> tuple[bool, str]:
+    """Check that the configured default model is registered in the model registry.
+
+    Retrieves the default model and provider from configuration and delegates
+    the availability check to the client holder.
+
+    Returns:
+        A tuple of (available, reason) where available is True if the default
+        model was found or no default model is configured, and reason describes
+        the outcome.
+    """
+    inference = configuration.inference
+    if (
+        inference is None
+        or not inference.default_model
+        or not inference.default_provider
+    ):
+        return True, "No default model configured"
+
+    expected_model_id = f"{inference.default_provider}/{inference.default_model}"
+
+    client_holder = AsyncLlamaStackClientHolder()
+    return await client_holder.check_model_available(expected_model_id)
+
+
 @router.get("/readiness", responses=get_readiness_responses)
 @authorize(Action.INFO)
 async def readiness_probe_get_method(
@@ -134,11 +160,21 @@ async def readiness_probe_get_method(
         unhealthy_provider_names = [p.provider_id for p in unhealthy_providers]
         reason = f"Providers not healthy: {', '.join(unhealthy_provider_names)}"
         response.status_code = status.HTTP_503_SERVICE_UNAVAILABLE
-    else:
-        ready = True
-        reason = "All providers are healthy"
+        return ReadinessResponse(
+            ready=ready, reason=reason, providers=unhealthy_providers
+        )
 
-    return ReadinessResponse(ready=ready, reason=reason, providers=unhealthy_providers)
+    # Check that the default model is registered in the model registry
+    model_available, model_reason = await check_default_model_available()
+    if not model_available:
+        response.status_code = status.HTTP_503_SERVICE_UNAVAILABLE
+        return ReadinessResponse(
+            ready=False, reason=model_reason, providers=unhealthy_providers
+        )
+
+    return ReadinessResponse(
+        ready=True, reason="All providers are healthy", providers=unhealthy_providers
+    )
 
 
 @router.get("/liveness", responses=get_liveness_responses)
--- a/src/client.py
+++ b/src/client.py
@@ -8,7 +8,7 @@ from typing import Optional
 import yaml
 from fastapi import HTTPException
 from llama_stack.core.library_client import AsyncLlamaStackAsLibraryClient
-from llama_stack_client import APIConnectionError, AsyncLlamaStackClient
+from llama_stack_client import APIConnectionError, APIStatusError, AsyncLlamaStackClient
 
 from configuration import configuration
 from llama_stack_configuration import YamlDumper, enrich_byok_rag, enrich_solr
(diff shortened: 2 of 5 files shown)


major:
@coderabbitai review

I've added a self-heal to the readiness probe in the hopes that we can heal pods if something breaks after startup.

major:
@coderabbitai review

I think I'm really really super duper done now, I promise (until the next revision). I'll bring you two baskets of carrots.

src/app/endpoints/health.py
```diff
@@ -95,6 +96,53 @@ async def get_providers_health_statuses() -> list[ProviderHealthStatus]:
...
+    default_provider = configuration.inference.default_provider
+
+    if not default_model or not default_provider:
+        return True, "No default model configured"
+
+    expected_model_id = f"{default_provider}/{default_model}"
+
+    try:
+        client = AsyncLlamaStackClientHolder().get_client()
+        models = await client.models.list()
```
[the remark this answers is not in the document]
major: Fixed in the current branch. The readiness path now catches RuntimeError while checking the default model and returns an unhealthy readiness result instead of letting /readiness crash; unit coverage includes the uninitialized-client case.

tisnik (approved):
LGTM

src/app/endpoints/health.py
```diff
@@ -95,6 +96,110 @@ async def get_providers_health_statuses() -> list[ProviderHealthStatus]:
         ]
 
 
+def _model_in_registry(models: list, expected_id: str) -> bool:
+    """Check if a model with the given ID exists in the model list."""
+    return any(model.id == expected_id for model in models)
+
+
+async def _reload_and_check_model(
```
asimurka: Think about moving this to `client.py` as a method of client holder or completely different decomposition based on the outcome from the comment below.

src/app/endpoints/health.py
```diff
@@ -95,6 +96,110 @@ async def get_providers_health_statuses() -> list[ProviderHealthStatus]:
...
+    Returns:
+        A tuple of (found, reason) where found is True if the model was
+        found after reloading the client.
+    """
+    logger.warning(
+        "Default model %s not found, attempting client reload",
+        expected_model_id,
+    )
+    try:
+        await client_holder.reload_library_client()
```
asimurka: What happens in server mode? 
Is there a reason for reloading the client? The reloading was originally designed for updated headers propagation. Models are statically configured on startup, so I think reloading is not necessary here.

src/app/endpoints/health.py
```diff
@@ -134,11 +239,21 @@ async def readiness_probe_get_method(
...
-        ready = True
-        reason = "All providers are healthy"
+        return ReadinessResponse(
+            ready=ready, reason=reason, providers=unhealthy_providers
+        )
+
+    # Check that the default model is registered in the model registry
+    model_available, model_reason = await check_default_model_available()
+    if not model_available:
+        response.status_code = status.HTTP_503_SERVICE_UNAVAILABLE
```
asimurka, line 170: No action required here, just for the record - 503 is not ideal here, although it is consistent with the above behavior. It does not conform to the overall architecture and is not documented in the Open API. It will very likely be refactored to a 200 status code in the future.

asimurka (changes requested):

src/app/endpoints/health.py
major: I'll take a look.

src/app/endpoints/health.py
major: I might be misunderstanding what _reloading the client_ means here. We have a situation like this:

1. Container starts with lightspeed-stack in it and llama-stack/ogx in library mode
2. Something goes wrong with the model registration via vertex AI (network blip, rate limit, API failure)
3. lightspeed-stack remains up with passing readiness/liveness checks
4. Any requests to endpoints get a response saying the default model cannot be found

That container keeps running and returning 404s until we can kill it, but we deploy in an environment where we do not have an access to kill a container. :| Should I be thinking about this another way?

src/app/endpoints/health.py
major: My goal was to try a reconnection once and then fail the check after. Does this make sense?

src/app/endpoints/health.py
asimurka: Yeah I get it now. Tbh I'm not sure what is the reload in library mode capable of. As I said, I created it for updating headers originally but I think it can serve also for this purpose... Let's keep it as it is ;)

src/app/endpoints/health.py
asimurka: Yeah the logic itself is fine and consistent with existing behavior (which is inconsistent with the rest of the project).
IMO `ReadinessResponse` should be accompanied with status code 200. But as I say this is out of scope for this PR. No action needed here.

src/app/endpoints/health.py
major: @asimurka I moved some of the logic around. Let me know if this makes more sense.

asimurka (approved):

PR #1703, opened 2026-05-08
LCORE- Unify logging configuration and setup

samdoran:
## Description

Major overhaul of the logging system:
- Unify the logging configuration from lightspeed-stack and uvicorn
- Change the time format to include milliseconds
- Pass the unified configuration into uvicorn so that all application logging works consistently
- Call `setup_logging()` early on so that logging is setup correctly.
- Call `setup_logging()` as appropriate when using `AsyncLlamaStackAsLibraryClient` to ensure the logging config in llama-stack is applied.
- Manually set the root logger name of `lightspeed_stack` so that the logging configuration is applied correctly.

The end result of these changes is that calling `get_logger(__name__)` results in a logger instance where `logger.info()` and other methods display log messages as expected and formatted consistently.

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Assisted-by: Claude
- Written by: me

None

## Testing
To run unit tests: `pytest tests/unit/test_log.py`
To test logging configuration without Rich logging: `LIGHTSPEED_STACK_DISABLE_RICH_HANDLER=1 lightspeed-stack` or `python src/lightspeed_stack.py`
To test logging configuration using the default format string: `lightspeed-stack` or `python src/lightspeed_stack.py`

## Summary by CodeRabbit

* **Refactor**
  * Centralized logging initialization for consistent behavior across app and server runner.
  * Standardized default logger naming for clearer log-source identification.
  * Improved plain-text log format (millisecond timestamps, reordered fields) and TTY-aware handler selection with an opt-out toggle.
  * Server runner now forwards the resolved logging configuration.

* **Tests**
  * Unit tests reset/init shared logging and use structured log capture for assertions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lightspeed-core/lightspeed-stack/pull/1703?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/log.py
+++ b/src/log.py
@@ -135,6 +135,7 @@ def build_logging_config() -> dict[t.Any, t.Any]:
             "%(asctime)s.%(msecs)03d %(levelprefix)s "
             '%(client_addr)s - "%(request_line)s" %(status_code)s'
         )
+        merged_config["formatters"]["access"]["datefmt"] = "%Y-%m-%d %H:%M:%S"
         merged_config["formatters"]["default"]["fmt"] = DEFAULT_LOG_FORMAT
         merged_config["formatters"]["default"]["datefmt"] = "%Y-%m-%d %H:%M:%S"
 
--- a/src/log.py
+++ b/src/log.py
@@ -57,9 +57,17 @@ def resolve_log_level() -> int:
 
 def get_logger(name: str) -> logging.Logger:
     """Create a common logger for all modules in this package."""
-    # Normally this is derived from the package name
-    return logging.getLogger(name)
-    # return logging.getLogger(f"{DEFAULT_LOGGER_NAME}.{name}")
+    # FIXME: Remove the need for this function.
+    #
+    # Normally this is derived from the package name (__name__).
+    #
+    # Since this program is sometimes called from from the entrypoint and
+    # sometimes called from src/lightspeed_stack.py, the value for __name__
+    # does not contain a consistent root value.
+    #
+    # How the application is installed and run needs to be streamlined so that
+    # __name__ provides the expected value in all cases.
+    return logging.getLogger(f"{DEFAULT_LOGGER_NAME}.{name}")
 
 
 @lru_cache
--- a/src/log.py
+++ b/src/log.py
@@ -56,6 +56,9 @@ def resolve_log_level() -> int:
     return validated_level
 
 
+def get_logger(file: str) -> logging.Logger:
+    return logging.getLogger(f"{DEFAULT_LOGGER_NAME}.{Path(file).stem}")
+
 
 @lru_cache
 def setup_logging() -> dict[t.Any, t.Any]:
--- a/src/log.py
+++ b/src/log.py
@@ -5,10 +5,12 @@ import logging.config
 import os
 import sys
 import typing as t
+from datetime import datetime
 from functools import lru_cache
 
 import uvicorn.config
 from pydantic.v1.utils import deep_update
+from rich.text import Text
 
 from constants import (
     DEFAULT_LOG_FORMAT,
@@ -17,6 +19,11 @@ from constants import (
     LIGHTSPEED_STACK_DISABLE_RICH_HANDLER_ENV_VAR,
     LIGHTSPEED_STACK_LOG_LEVEL_ENV_VAR,
 )
+
+
+def _ms_time_format(dt: datetime) -> Text:
+    """Format datetime object with zero padded milliseconds."""
+    return Text(dt.strftime("%Y-%m-%d %H:%M:%S.") + f"{dt.microsecond // 1000:03d}")
 
 
 def resolve_log_level() -> int:
@@ -87,7 +94,7 @@ def setup_logging() -> dict[t.Any, t.Any]:
             "rich": {
                 "()": "rich.logging.RichHandler",
                 "show_time": True,
-                "log_time_format": "%Y-%m-%d %H:%M:%S.%f",
+                "log_time_format": _ms_time_format,
                 "level": log_level,
             },
         },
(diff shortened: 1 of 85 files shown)


tisnik (changes requested):
please try to figure out why unit tests fail. LGTM in overall

samdoran:
Yup, I'm working on the unit test failures.

samdoran:
The problem with the tests was `caplog` relying on propagation of log messages. I had to change the fixture that configures logging for the tests to ensure propagation is enabled (it is not in the new default config) and then restore state after each test.

samdoran:
I made a few more changes, the most significant is writing our own recursive function to merge mappings. It's a simpler implementation than `deep_update` but similar.

src/log.py
```diff
@@ -50,62 +76,70 @@ def resolve_log_level() -> int:
...
+            "%(asctime)s.%(msecs)03d %(levelprefix)s "
+            '%(client_addr)s - "%(request_line)s" %(status_code)s'
+        )
+        merged_config["formatters"]["default"]["fmt"] = DEFAULT_LOG_FORMAT
+        merged_config["formatters"]["default"]["datefmt"] = "%Y-%m-%d %H:%M:%S"
 
-    # Skip reconfiguration if logger already has handlers from a prior call
-    if logger.handlers:
-        return logger
+    logging.config.dictConfig(merged_config)
```
Lifto: Tested this locally with `use_as_library_client=True`. The config applied here works correctly at startup — all loggers (including uvicorn) pick up the new format:

```
2026-05-21 21:12:03.661 INFO:     Lightspeed Core Stack startup  [lightspeed_stack.__main__:94]
2026-05-21 21:12:06.065 INFO:     Started server process [1]  [uvicorn.error:92]
```

However, during ASGI startup the llama_stack library client calls its own `logging.config.dictConfig()`, which overwrites the uvicorn logger formatters. After that point, uvicorn switches to llama_stack's format:

```
INFO     2026-05-21 21:12:08,869 uvicorn.error:62 server: Application startup complete.
INFO     2026-05-21 21:20:38,289 uvicorn.access:481 server: 127.0.0.1:56220 - "GET /api/lightspeed/health/live HTTP/1.1"
```

The `lightspeed_stack.*` loggers survive because llama_stack's config doesn't mention them and `disable_existing_loggers` is `False`. But uvicorn's loggers get reconfigured, so you end up with two formats interleaved in the same output.

**Potential fix**: re-apply after llama_stack finishes initializing. Since `setup_logging()` is `@lru_cache`'d, a second `logging.config.dictConfig(setup_logging())` call at the right point would return the cached dict and overwrite llama_stack's config.

src/runners/uvicorn.py
```diff
@@ -30,6 +37,7 @@ def start_uvicorn(configuration: ServiceConfiguration) -> None:
         host=configuration.host,
         port=configuration.port,
         workers=configuration.workers,
+        log_config=log_config,
         log_level=log_level,
```
Lifto, line 41: Related to the `use_colors=True` a few lines below: in a non-TTY environment (containers), that forces ANSI escape codes into the log stream. The `%(levelprefix)s` variable in uvicorn's `DefaultFormatter` wraps level names in color codes — `\033[36mDEBUG\033[0m`, `\033[33mWARNING\033[0m`, etc. — which render as raw `[36m`/`[0m` in collected logs.

`setup_logging()` correctly sets `use_colors: null` on the formatters, which auto-detects via `stream.isatty()`. But uvicorn's `Config.configure_logging()` checks `if self.use_colors in (True, False)` and overrides the formatters — so `True` undoes the auto-detect.

Changing `use_colors=True` to `use_colors=None` would let the formatters keep their auto-detection.

samdoran: I'm not sure why it is there currently but I'll just omit that. Thanks for finding that.

src/log.py
samdoran: This is a down side to logging being a global system in Python. I'll dig into this a bit more and see if I can find a better place to initialize the logging.

samdoran: It looks like ~llama-stack~ `ogx` is doing the [_exact same thing_](https://github.com/ogx-ai/ogx/blob/41c94c49dc252c4fe8029d2bf3972fbe2290722d/src/ogx/log.py#L435-L482) I'm doing in this PR: they have a function to setup logging that is called early that merges the logging config with that of `uvicorn`. I need to rectify that with these changes.

samdoran: Thinking about this more and looking through the code I see why this is only an issue when using `ogx` as a library: The `__init__` of `AsyncLlamaStackAsLibraryClient` has an explicit call to `setup_logging` in `ogx`. Since that runs _after_ the `setup_logging` here and logging is a global, the last config wins.

samdoran: I verified this by instrumenting the calls to `setup_logging()` in both places.

samdoran: Ok, I came up with a clean solution for this. I also decided to separate configuration building from logging configuration. That was a messy implementation.

samdoran:
This is ready for final review now.

Lifto:
I tried it out again and the logging looks good

samdoran:
The test failure on [E2E Tests / E2E: server mode / ci / group 2 (pull_request_target)](https://github.com/lightspeed-core/lightspeed-stack/actions/runs/28174312932/job/83446386354?pr=1703) seems like the job just hung and it needs to be rerun.

samdoran:
The failed test passed now. Just need to trigger the Konflux test to make sure it builds ok.

tisnik (approved):
LGTM

PR #1736, opened 2026-05-13
LCORE-2072: Add SkillsConfiguration model to config file

anik120:
## Description

Adds the `SkillsConfiguration` Pydantic model to enable configuring skill directory paths in `lightspeed-stack.yaml`. This is the first step in implementing Agent Skills support.

For more info, refer to `docs/design/agent-skills/agent-skills.md`.

**Scope**: This PR adds only the configuration model. Runtime skill loading (`load_skills()`, frontmatter parsing, tool registration) will be implemented in follow-up commits.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Assisted-by: Claude

- Related Issue #
- Closes #

## Testing

## Summary by CodeRabbit

* **New Features**
  * Add configurable "skills" support allowing multiple skill directory paths and optional skills config in service settings
  * Example configuration added showing library-client mode and local user data collection for feedback and transcripts

* **Documentation**
  * API schema updated to include the new skills configuration object and descriptive guidance

* **Tests**
  * Unit tests added to validate skills configuration parsing and path handling

[Review Change Stack](https://app.coderabbit.ai/change-stack/lightspeed-core/lightspeed-stack/pull/1736?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/examples/lightspeed-stack-skills.yaml
+++ b/examples/lightspeed-stack-skills.yaml
@@ -1,0 +1,31 @@
+name: Lightspeed Core Service (LCS) with Skills
+service:
+  host: localhost
+  port: 8080
+  auth_enabled: false
+  workers: 1
+  color_log: true
+  access_log: true
+llama_stack:
+  use_as_library_client: true
+  library_client_config_path: run.yaml
+user_data_collection:
+  feedback_enabled: true
+  feedback_storage: "/tmp/data/feedback"
+  transcripts_enabled: true
+  transcripts_storage: "/tmp/data/transcripts"
+authentication:
+  module: "noop"
+# Agent skills configuration
+# Skills provide domain-specific instructions and reference materials
+# that the LLM can load on demand when relevant to the current task
+skills:
+  paths:
+    # Option A: Directory containing multiple skill subdirectories
+    # Each subdirectory must contain a SKILL.md file
+    - "/var/skills/"
+
+    # Option B: Individual skill paths for fine-grained control
+    # - "/var/skills/openshift-troubleshooting/"
+    # - "/var/skills/code-review/"
+    # - "/opt/custom-skills/deployment-guide/"
--- a/src/models/config.py
+++ b/src/models/config.py
@@ -1944,6 +1944,26 @@ class AzureEntraIdConfiguration(ConfigurationBase):
     )
 
 
+class SkillsConfiguration(ConfigurationBase):
+    """Agent skills configuration.
+
+    Specifies paths to skill directories. Skill metadata (name, description)
+    is read from SKILL.md frontmatter at startup.
+
+    Each path can point to either:
+    - A directory containing a SKILL.md file (single skill)
+    - A directory containing subdirectories with SKILL.md files (multiple skills)
+
+    Paths are validated at startup to ensure they exist and contain valid SKILL.md files.
+    """
+
+    paths: list[Path] = Field(
+        default_factory=list,
+        title="Skill paths",
+        description="Paths to skill directories or directories containing skill subdirectories.",
+    )
+
+
 class Configuration(ConfigurationBase):
     """Global service configuration."""
 
@@ -2108,6 +2128,12 @@ class Configuration(ConfigurationBase):
         default_factory=RerankerConfiguration,
         title="Reranker configuration",
         description="Configuration for neural reranking of RAG chunks using cross-encoder.",
+    )
+
+    skills: Optional[SkillsConfiguration] = Field(
+        default=None,
+        title="Agent skills",
+        description="Agent skills configuration. Specifies paths to skill directories.",
     )
 
     @model_validator(mode="after")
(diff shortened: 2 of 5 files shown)


tests/unit/models/config/test_skills_configuration.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def test_mixed_absolute_and_relative_paths(self) -> None:
+        """Test that both absolute and relative paths can be mixed."""
+        config = SkillsConfiguration(
+            paths=["/var/skills", "./local-skills", "/opt/skills"]
+        )
+        assert len(config.paths) == 3
+        assert "/var/skills" in config.paths
+        assert "./local-skills" in config.paths
+        assert "/opt/skills" in config.paths
```
anik120: @radofuchs what do you think about this test? From the code's perspective, absolute/relative doesn't matter. It's just "str" at the end of the day - but I added this coz it signals that we'll work with both absolute and relative path - ie this is my attempt at BDD :)

examples/lightspeed-stack-skills.yaml
```diff
@@ -0,0 +1,31 @@
...
+  feedback_storage: "/tmp/data/feedback"
+  transcripts_enabled: true
+  transcripts_storage: "/tmp/data/transcripts"
+authentication:
+  module: "noop"
+# Agent skills configuration
+# Skills provide domain-specific instructions and reference materials
+# that the LLM can load on demand when relevant to the current task
+skills:
+  paths:
```
jrobertboos, line 23: is there a reason that we need the paths? It would make sense if we had other data fields under skills but if paths is the only one then I think skills can just be a list. wdyt?

anik120: ~Yea I think that makes total sense. I was going off of what's here https://github.com/lightspeed-core/lightspeed-stack/blob/main/docs/design/agent-skills/agent-skills.md#configuration~ 


~but I do remember this discussion and AFAIR we'd decided we don't need path. I'll update the design doc too~



Based on further discussions: 

It's a little weird to look at now, but the current layout is the safest approach - I can't think of anything else that might be needed under the skills tab (eg, settings etc), but keeping it tabbe-ed future proofs it so keeping it as is

src/models/config.py
```diff
@@ -1944,6 +1944,26 @@ class AzureEntraIdConfiguration(ConfigurationBase):
...
+    is read from SKILL.md frontmatter at startup.
+
+    Each path can point to either:
+    - A directory containing a SKILL.md file (single skill)
+    - A directory containing subdirectories with SKILL.md files (multiple skills)
+
+    Paths are validated at startup to ensure they exist and contain valid SKILL.md files.
+    """
+
+    paths: list[str] = Field(
```
tisnik: `list[Path]` perhaps

tests/unit/models/config/test_dump_configuration.py
```diff
@@ -235,6 +235,7 @@ def test_dump_configuration(tmp_path: Path) -> None:
                 "enabled": False,
```
tisnik, line 236: you can enhance the config file used in dump tests, and then it will be possible to check for actual skills path processing

src/models/config.py
```diff
@@ -1944,6 +1944,26 @@ class AzureEntraIdConfiguration(ConfigurationBase):
...
+        default_factory=list,
+        title="Skill paths",
+        description="Paths to skill directories or directories containing skill subdirectories.",
+    )
+
```
tisnik, line 1965: prob. check if the directory is accessible. It might be too early to do it during config loading, but probably better from UX perspective (fail early...). WDYT?

tisnik (changes requested):
One or two nits (str->Path + directory checks), LGTM in overall,

src/models/config.py
anik120: Changed, PTAL

tests/unit/models/config/test_dump_configuration.py
anik120: Added a `test_dump_configuration_with_skills`, is that what you meant?

src/models/config.py
anik120: I'm leaning towards keeping filesystem validation out of the Pedantic config layer - eg in containerized environments volumes may not be mounted yet when the config is parsed. 

It also makes unit testing harder since we have to create actual directories. 


Fwiw,  `load_skills()` will need to have this validation function, and that looks like the earliest we can validate safely?

tisnik (approved):
LGTM

jrobertboos (approved):
LGTM

PR #1805, opened 2026-05-27
RSPEED-3116: explicitly set type='mcp' on InputToolMCP constructions

major:
## Description

MCP-enabled requests in library-client mode fail with `union_tag_not_found` on `CreateResponseRequest.tools.0`. The input dict is missing the `type` key.

- `InputToolMCP(...)` is built without passing `type='mcp'`, so Pydantic fills it from the parent class default.
- The llama-stack client SDK serializes via `model_dump(exclude_unset=True)`, which drops defaulted fields, so `type` never leaves our process.
- In library-client mode, `library_client._convert_body` revalidates the dict against the discriminated union and blows up.

Fix: pass `type="mcp"` at both `InputToolMCP` construction sites in `src/utils/responses.py` so the field is set and survives `exclude_unset`.

Note: redeclaring `type: Literal['mcp'] = 'mcp'` on the subclass does not fix it. Defaulted is still unset under Pydantic v2.

## Type of change

- [x] Bug fix
- [x] Unit tests improvement

## Tools used to create PR

- Assisted-by: Claude (opencode), CodeRabbit (local, 0 findings)
- Generated by: N/A

- Refs: RSPEED-3116

## Testing

- `uv run make format` and `uv run make verify` clean.
- New `TestInputToolMCPTypeDiscriminator` covers both call sites plus the underlying `model_dump(exclude_unset=True)` behavior.
- Sanity check: with the source fix reverted, the two call-site tests fail with `assert None == 'mcp'`. With the fix in place, all three pass.

### DIFF
--- a/src/utils/responses.py
+++ b/src/utils/responses.py
@@ -743,6 +743,13 @@ async def get_mcp_tools(
         )
         tools.append(
             InputToolMCP(
+                # Pass type explicitly: the llama-stack client serializes pydantic
+                # instances with model_dump(exclude_unset=True), which strips fields
+                # filled from defaults. Without an explicit value here, the 'type'
+                # discriminator is dropped before reaching CreateResponseRequest,
+                # causing a union_tag_not_found validation error in library-client
+                # mode. See RSPEED-3116.
+                type="mcp",
                 server_label=mcp_server.name,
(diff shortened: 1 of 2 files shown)


tisnik (approved):
LGTM

PR #1844, opened 2026-06-03
RSPEED-3082: add build-nudge-files annotation for cross-repo nudging

Lifto:
## What

Add `build.appstudio.openshift.io/build-nudge-files` annotation to the push PipelineRun, targeting `openshift/lightspeed-stack.yml` in lscore-deploy.

## Why

The Konflux nudge controller's default file pattern (`.*Dockerfile.*, .*.yaml, .*Containerfile.*`) only matches `.yaml` files. Since lscore-deploy uses `openshift/lightspeed-stack.yml` (note: `.yml`), the nudge controller silently skips it — no MR is opened to bump the image digest.

This annotation explicitly tells the nudge controller which file to scan, bypassing the default pattern entirely.

## Tickets

- [RSPEED-3082](https://redhat.atlassian.net/browse/RSPEED-3082)

## Testing

Once merged, the push pipeline triggered by this merge will include the annotation. The nudge controller should then open an MR in lscore-deploy updating the image digest in `openshift/lightspeed-stack.yml`.

## Summary by CodeRabbit

* **Chores**
  * Added a new metadata annotation to the build pipeline configuration.
  * Updated pipeline metadata to include a reference to a stack configuration file.
  * Minor metadata-only change; no functional behavior changes expected.

### DIFF
--- a/.tekton/lightspeed-stack-push.yaml
+++ b/.tekton/lightspeed-stack-push.yaml
@@ -2,6 +2,10 @@ apiVersion: tekton.dev/v1
 kind: PipelineRun
 metadata:
   annotations:
+    # Cross-repo nudge: after each push build, the Konflux nudge controller
+    # opens an MR in lscore-deploy (gitlab.cee.redhat.com/rhel-lightspeed/lscore-deploy)
+    # to bump the image digest in openshift/lightspeed-stack.yml. See RSPEED-3082.
+    build.appstudio.openshift.io/build-nudge-files: "openshift/lightspeed-stack.yml"
     build.appstudio.openshift.io/repo: https://github.com/lightspeed-core/lightspeed-stack?rev={{revision}}
     build.appstudio.redhat.com/commit_sha: '{{revision}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'


samdoran:
Wouldn't it be more straightforward to move the deployment config from `lscore-deploy` to this repo? Then we would be able to auto-deploy images to stage once changes are merged to this repo and the images are built.

Adding a nudge to this repo that references a file that doesn't exist anywhere in this repo may be confusing to future contributors.

Lifto:
> Wouldn't it be more straightforward to move the deployment config from `lscore-deploy` to this repo? Then we would be able to auto-deploy images to stage once changes are merged to this repo and the images are built.
> 
> Adding a nudge to this repo that references a file that doesn't exist anywhere in this repo may be confusing to future contributors.

Sam, thanks for your comment. Regarding the confusion I added a comment to the yml. 

Some reasoning for why I want to do it this way:
* rls-backend is a separate project from lightspeed-stack
* lscore-deploy is the canonical home for rls-backend's deployment config
    * I need independent control over rls-backend's deploy because I'm responsible for rls-backend (or rhel-lightspeed team is, anyway.)
    * I'm not a maintainer on lightspeed-stack. Moving the template here puts deployment config behind a review gate I don't control
* Cross-repo nudging is not a hack, it's a proper Konflux feature for such situations.
* RSPEED-3082 considers such colocation "Alternative: Option A" and chose cross-repo nudging instead

Lifto:
@samdoran There may be another option that lets us remove this change. Konflux has a default pattern that looks for files to do the digest replacement in, but Konflux's default file patterns only match .yaml, not .yml. We can override that pattern via a namespace ConfigMap in lightspeed-core-tenant so it also picks up .yml files. In that case we would not need this change, the cross-repo wiring would live entirely in Konflux cluster config. Do you think that's a better way?

(I have not tested that, btw. Claude tells me we have this option.)

tisnik (approved):
LGTM

samdoran:
> lscore-deploy is the canonical home for rls-backend's deployment config

That's the fundamental problem: app-interface expects the deployment config to live in the same repo as the code being deployed. That allows the container image tag to match the deployment config hash for automatic deployment. Fighting that design, limiting as it is, adds complexity.

I agree it would be odd to have our deployment config here, but that arrangement would align best with what app-interface recommends.

> I need independent control over rls-backend's deploy because I'm responsible for rls-backend (or rhel-lightspeed team is, anyway.)
> I'm not a maintainer on lightspeed-stack. Moving the template here puts deployment config behind a review gate I don't control

These two things are easily solved with a [codeowners](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) file.

> Cross-repo nudging is not a hack, it's a proper Konflux feature for such situations.

Nudging is a very useful feature. This is the incorrect use of nudging. While this change does seem to work, it crosses many boundaries that were not meant to be crossed.

[Nudging](https://konflux-ci.dev/docs/building/component-nudges) is meant for updating images used by components of the same application. A relationship is defined between the components of an application in Konflux within the same tenant. If component B uses component A in a `FROM` clause and component A changes, then component B needs to be rebuilt.


The simplest solution is to move the config from `lscore-deploy` to this repo.

PR #1885, opened 2026-06-09
LCORE-2443: Simplify OKP RAG setup with git submodules and auto-enrichment

anik120:
## Description

Rewire OKP installation for new container set up. As a byproduct, replace manual provider installation with git submodules. Users no longer need to 
1. Clone lightspeed-providers separately or set EXTERNAL_PROVIDERS_DIR.
2. Worry about llama-stack configuration (it is now an implementation detail hidden from user)

Changes:
- Add lightspeed-providers as submodule
- Bake providers into container via PYTHONPATH
- Automatic llama-stack config enrichment - users no longer have to worry about llama-stack
- Update CI/CD to fetch submodules
- Simplify OKP guide documentation

Setup: clone with --recursive, configure okp section, run make run.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Related Issue #
- Closes #

## Testing

## Summary by CodeRabbit

* **New Features**
  * Added first-class support for repository submodules to include external providers.

* **Documentation**
  * Installation and deployment guides updated with submodule cloning and new setup steps.

* **Chores**
  * CI/CD workflows updated to fetch submodules during build and test runs.
  * Container and runtime setup adjusted to include provider files at build/runtime.

* **Bug Fixes**
  * OKP URL environment variables now resolve correctly before use.

### DIFF
--- a/.github/workflows/e2e_tests_lightspeed_evaluation.yaml
+++ b/.github/workflows/e2e_tests_lightspeed_evaluation.yaml
@@ -30,6 +30,9 @@ jobs:
           # Don’t keep credentials when running untrusted PR code under PR_TARGET.
           persist-credentials: ${{ github.event_name != 'pull_request_target' }}
 
+          # Fetch submodules (required for lightspeed-providers)
+          submodules: 'recursive'
+
       - name: Verify actual git checkout result
         run: |
           echo "=== Git Status After Checkout ==="
--- a/src/llama_stack_configuration.py
+++ b/src/llama_stack_configuration.py
@@ -433,9 +433,11 @@ def enrich_solr(  # pylint: disable=too-many-locals
     )
 
     rhokp_raw = okp_config.get("rhokp_url")
-    base_url = (
+    base_url_raw = (
         str(rhokp_raw) if rhokp_raw is not None else constants.RH_SERVER_OKP_DEFAULT_URL
     )
+    # Resolve environment variables in the URL (e.g., ${env.RH_SERVER_OKP})
+    base_url = replace_env_vars(base_url_raw)
     solr_url = urljoin(base_url, "/solr")
 
     logger.info("Enriching Llama Stack config with OKP")
--- a/.github/workflows/build_and_push_dev.yaml
+++ b/.github/workflows/build_and_push_dev.yaml
@@ -26,6 +26,9 @@ jobs:
           sudo apt install -y buildah qemu-user-static
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          # Fetch submodules (required for lightspeed-providers)
+          submodules: 'recursive'
       - name: Create dev image tag
         run: |
           echo "DEV_TAG=dev-$(date +%Y%m%d)-$(git rev-parse --short HEAD)" >> $GITHUB_ENV
--- a/.github/workflows/build_and_push_release.yaml
+++ b/.github/workflows/build_and_push_release.yaml
@@ -28,6 +28,9 @@ jobs:
           sudo apt install -y buildah qemu-user-static
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          # Fetch submodules (required for lightspeed-providers)
+          submodules: 'recursive'
       - name: Build image with Buildah
         id: build_image
         uses: redhat-actions/buildah-build@v2
--- a/.github/workflows/build_pr.yaml
+++ b/.github/workflows/build_pr.yaml
@@ -24,6 +24,9 @@ jobs:
           sudo apt install -y buildah qemu-user-static
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          # Fetch submodules (required for lightspeed-providers)
+          submodules: 'recursive'
       - name: Build image with Buildah
         id: build_image
         uses: redhat-actions/buildah-build@v2
--- a/.github/workflows/e2e_tests.yaml
+++ b/.github/workflows/e2e_tests.yaml
@@ -37,6 +37,9 @@ jobs:
           # Don’t keep credentials when running untrusted PR code under PR_TARGET.
           persist-credentials: ${{ github.event_name != 'pull_request_target' }}
 
+          # Fetch submodules (required for lightspeed-providers)
+          submodules: ‘recursive’
+
       - name: Verify actual git checkout result
         run: |
           echo "=== Git Status After Checkout ==="
--- a/.github/workflows/e2e_tests_providers.yaml
+++ b/.github/workflows/e2e_tests_providers.yaml
@@ -62,6 +62,9 @@ jobs:
           # Don’t keep credentials when running untrusted PR code under PR_TARGET.
           persist-credentials: ${{ github.event_name != 'pull_request_target' }}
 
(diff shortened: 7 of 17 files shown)


src/llama_stack_configuration.py
```diff
@@ -433,9 +433,11 @@ def enrich_solr(  # pylint: disable=too-many-locals
     )
 
     rhokp_raw = okp_config.get("rhokp_url")
-    base_url = (
+    base_url_raw = (
         str(rhokp_raw) if rhokp_raw is not None else constants.RH_SERVER_OKP_DEFAULT_URL
     )
+    # Resolve environment variables in the URL (e.g., ${env.RH_SERVER_OKP})
+    base_url = replace_env_vars(base_url_raw)
```
Jazzcort, line 440: Was the default value syntax (${env.FOO:=default}) working before for `EXTERNAL_PROVIDER_DIR`? Seems like it would not support that syntax without calling `replace_env_vars(...)` 😆

Jazzcort (approved):
Like what I mentioned in the chat. If we decide to pull the providers for our users instead of having them setting these up, hardcoding the directory in `run.yaml` seems like a right move. But we need to clean up those references for `EXTERNAL_PROPVIDER_DIR` from all of the docs so it won't cause the confusion in the future. 

Image

Here is the screenshot of those reference I found. 😁 

Everything else looks good to me! We can always fix the workflow if it breaks but I think it will be fine.

src/llama_stack_configuration.py
anik120: Great catch 😄  `${env.EXTERNAL_PROVIDERS_DIR}` in `run.yaml` was not being resolved by the enrichment script - llama-stack itself was resolving it. The enrichment script never touched that variable, just passed it directly to llama-stack, which in turn was reading the enriched `run.yaml` and llama-stack's own config parser was resolving all ${env.Var} patterns. 

So without `replace_env_vars()`, we'd be doing:

```
base_url = "${env.RH_SERVER_OKP}"  # literal string, not resolved!
solr_url = urljoin("${env.RH_SERVER_OKP}", "/solr")
# Result: solr_url = "${env.RH_SERVER_OKP}/solr" (malformed)
```

So yes, the default value syntax ${env.FOO:=default} wouldn't have worked in the enrichment script before `replace_env_vars()` was added. It only worked for values that passed straight through to llama-stack's parser! 😆

anik120:
> Like what I mentioned in the chat. If we decide to pull the providers for our users instead of having them setting these up, hardcoding the directory in `run.yaml` seems like a right move. But we need to clean up those references for `EXTERNAL_PROPVIDER_DIR` from all of the docs so it won't cause the confusion in the future.
> 
> Image
> Here is the screenshot of those reference I found. 😁
> 
> Everything else looks good to me! We can always fix the workflow if it breaks but I think it will be fine.

Thanks for the review @Jazzcort and good catch. PTAL at https://github.com/lightspeed-core/lightspeed-stack/pull/1885/commits/eb9e5c9191354c665a09f359753b36f95a872d23 

Files Updated:
     
1. docs/container_orchestration.md - Removed from "Other Configuration" section
2. docker-compose.yaml - Removed environment variable
  
Remaining References:
  
- `providers/README.md` - This is in the providers submodule (external repo, so we should go clean that up in the [providers](https://github.com/lightspeed-core/lightspeed-providers/tree/main) repo)
- `docs/design/llama-stack-config-merge/ `- Design/spike documents (historical context, not user-facing docs)

Jazzcort (approved):
LGTM! 😁

src/llama_stack_configuration.py
Jazzcort: Thanks for the deep dive here! I was just suspecting because I saw the default syntax was using on `EXTERNAL_PROVIDERS_DIR` somewhere. 😆

.github/workflows/e2e_tests_lightspeed_evaluation.yaml
```diff
@@ -30,6 +30,9 @@ jobs:
           # Don’t keep credentials when running untrusted PR code under PR_TARGET.
           persist-credentials: ${{ github.event_name != 'pull_request_target' }}
 
+          # Fetch submodules (required for lightspeed-providers)
+          submodules: 'recursive'
+
```
Anxhela21, line 35: why are we changing e2e workflow jobs in this PR?

anik120: We're introducing a new way of pulling in the providers repo - e2e jobs need to adopt that

anik120: Also see the description of the PR that has more info

Anxhela21: Wasn't the end goal to remove providers?

Anxhela21: there doesn't seem to be an issue or epic associated with this PR. Where is the motivation coming from?

Anxhela21: A git submodule would make lightspeed-stack depend on a specific commit of lightspeed-providers, not automatically on “latest main” right?

anik120: >there doesn't seem to be an issue or epic associated with this PR. 

Isn't LCORE-2443 the associated issue, and its parent, the epic? 

> Where is the motivation coming from?

You're in the #team-collab-okp-lcore channel right? https://redhat-internal.slack.com/archives/C092RKNEQ81/p1780415148953239?thread_ts=1780406932.917899&cid=C092RKNEQ81

> A git submodule would make lightspeed-stack depend on a specific commit of lightspeed-providers, not automatically on “latest main” right?

A quick google search would land you to https://git-scm.com/book/en/v2/Git-Tools-Submodules - which would give you the below info of how it works: 
     
 1. Initial setup: When you add the submodule with git submodule add, lightspeed-stack records the current commit SHA of lightspeed-providers in its tree
2. Cloning behavior: When someone clones lightspeed-stack and runs git submodule update --init, they get that exact pinned commit, NOT the latest from lightspeed-providers
3. Updating the submodule: To use a newer version of lightspeed-providers, you must explicitly:

```  
cd providers/
git pull origin main  # or checkout specific commit
cd ..
git add providers/
git commit -m "Update providers submodule to commit abc123"
```

You can check which commit is currently pinned with:

```
$ git ls-tree HEAD providers/
160000 commit 9ab872996b1b95448a9332bc3b468fe27b4ee2e6	providers
```

anik120: Btw, yes that means manual maintenance is required - need to periodically update the submodule reference.
If forgotten, it could lead to staleness and we could be using outdated providers till someone remembers.

tisnik (approved):
LGTM

PR #1895, opened 2026-06-10
Speed up local dev loop by reusing llama-stack containers

Jazzcort:
## Description

Previously, `make run` always rebuilt the container image from scratch.
Now start-llama-stack-container reuses running/stopped containers and
only builds when no image exists. Use FRESH=true to force a rebuild.

Also extracts ensure-container-runtime as a shared prerequisite and
inlines the removed run-stack target.

---

"make run-stack works for skipping rebuilds, but having to manually start the container is a bit of a friction point. I’ve patched make run so it stays true to being our 'one-command' setup. I know we’re moving away from llama-stack eventually, but I figured we might as well make it as smooth as possible while we're still using it! 😁"

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Assisted-by: claude-opus-4.6

- Related Issue #
- Closes #

## Testing
1. If the dependencies are not changed, `make run` should just rebuild the app image and should take less time than before.
2. If the dependencies are changed, `make run` should rebuild both deps and app images.
3. There shouldn't be any dangling `none:none` images after executing `make run` for multiple times.

## Summary by CodeRabbit

* **Chores**
  * Added dependency-change detection to avoid unnecessary rebuilds.
  * Introduced a dedicated deps image and ensures stale containers/images are removed before rebuild/start.
  * Clean routine now removes the deps image and its stored hash.
  * .gitignore updated to exclude the auto-generated deps hash file.

* **Refactor**
  * Converted container build to a multi-stage layout for smaller runtime images and improved layer caching.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,27 +16,27 @@ LLAMA_STACK_CONTAINER_NAME ?= lightspeed-llama-stack
 LLAMA_STACK_IMAGE ?= lightspeed-llama-stack:local
 LLAMA_STACK_PORT ?= 8321
 CONTAINER_RUNTIME ?= $(shell command -v podman 2>/dev/null || command -v docker 2>/dev/null)
+FRESH ?= false
 
-.PHONY: run run-stack build-llama-stack-image remove-llama-stack-container stop-llama-stack-container start-llama-stack-container wait-for-llama-stack-health clean-llama-stack
-
-run-stack: ## Run lightspeed-stack directly, without building dependent service/s
-	uv run src/lightspeed_stack.py -c $(CONFIG)
+.PHONY: run ensure-container-runtime build-llama-stack-image stop-llama-stack-container remove-llama-stack-container start-llama-stack-container wait-for-llama-stack-health clean-llama-stack
 
 run: start-llama-stack-container ## Run the service locally with dependent services
 	@echo "Starting Lightspeed Core Stack..."
-	@trap 'echo ""; echo "Stopping services..."; $(MAKE) stop-llama-stack-container' EXIT INT TERM; \
-	$(MAKE) run-stack
+	@trap 'echo ""; echo "Stopping services..."; $(MAKE) stop-llama-stack-container' INT TERM; \
+	uv run src/lightspeed_stack.py -c $(CONFIG)
 
-build-llama-stack-image: remove-llama-stack-container ## Build llama-stack container image
-	@echo "Building llama-stack container image..."
+ensure-container-runtime:
 	@if [ -z "$(CONTAINER_RUNTIME)" ]; then \
 		echo "ERROR: No container runtime found. Install podman or docker."; \
 		exit 1; \
 	fi
+
+build-llama-stack-image: ensure-container-runtime clean-llama-stack ## Build llama-stack container image
+	@echo "Building llama-stack container image..."
 	$(CONTAINER_RUNTIME) build -f deploy/llama-stack/test.containerfile -t $(LLAMA_STACK_IMAGE) .
 
-stop-llama-stack-container: ## Gracefully stop llama-stack container
-	@if [ -n "$(CONTAINER_RUNTIME)" ] && $(CONTAINER_RUNTIME) inspect $(LLAMA_STACK_CONTAINER_NAME) >/dev/null 2>&1; then \
+stop-llama-stack-container: ensure-container-runtime ## Gracefully stop llama-stack container
+	@if $(CONTAINER_RUNTIME) inspect $(LLAMA_STACK_CONTAINER_NAME) >/dev/null 2>&1; then \
 		echo "Stopping llama-stack container (timeout: 10s)..."; \
 		if $(CONTAINER_RUNTIME) stop -t 10 $(LLAMA_STACK_CONTAINER_NAME) 2>/dev/null; then \
 			echo "✓ Container stopped gracefully"; \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -27,13 +33,44 @@ run: start-llama-stack-container ## Run the service locally with dependent servi
...
+	@echo "Building llama-stack deps image..."
 	@if [ -z "$(CONTAINER_RUNTIME)" ]; then \
 		echo "ERROR: No container runtime found. Install podman or docker."; \
 		exit 1; \
 	fi
-	$(CONTAINER_RUNTIME) build -f deploy/llama-stack/test.containerfile -t $(LLAMA_STACK_IMAGE) .
+	@if $(CONTAINER_RUNTIME) image inspect $(LLAMA_STACK_DEPS_IMAGE) >/dev/null 2>&1; then \
+		echo "Removing existing deps image to avoid dangling images..."; \
+		$(CONTAINER_RUNTIME) rmi $(LLAMA_STACK_DEPS_IMAGE); \
+	fi
```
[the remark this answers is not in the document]
Jazzcort: Why do we need a fallback image if we're building a developer build for testing? 🤔 If the build fails, we fix the problem and build it again. Cleaning those dangling images is just too tedious and they are huge.

Jazzcort:
This patch breaks the CI process in E2E server mode test. Should be a easy fix. I'll fix it later. 😁

tests/integration/container_lifecycle/test_container_lifecycle.py
```diff
@@ -161,7 +161,13 @@ def test_build_is_idempotent_via_image_id(self, container_runtime):
...
         # Trigger the first build
         subprocess.run(
-            ["make", "build-llama-stack-image"],
+            [
+                container_runtime,
+                "build",
+                "-f",
+                "deploy/llama-stack/test.containerfile",
+                "lightspeed-llama-stack",
+            ],
```
anik120, line 170: Using make targets ensures we test what users use - there's a chance in the future that the make target could deviate from what's in here, so if we don't test the make targets directly, we'll miss something.

Makefile
```diff
@@ -48,17 +48,32 @@ stop-llama-stack-container: ## Gracefully stop llama-stack container
...
+			echo "Starting existing container..."; \
+			$(CONTAINER_RUNTIME) start $(LLAMA_STACK_CONTAINER_NAME); \
+		fi; \
+		$(MAKE) wait-for-llama-stack-health; \
+		exit 0; \
+	elif ! $(CONTAINER_RUNTIME) image inspect $(LLAMA_STACK_IMAGE) >/dev/null 2>&1; then \
+		echo "Image not found, building..."; \
+		$(MAKE) build-llama-stack-image; \
+	fi; \
+	echo "Starting llama-stack container..."; \
```
anik120, line 76: Agree with the principal of this 👍🏽 

What do you think about moving this logic to a bash script instead of keeping it in the makefile? 


That way the Makefile is clean and easy to read/debug in the future

```
run-llama-stack:
	@CONTAINER_RUNTIME=$(CONTAINER_RUNTIME) \
	 LLAMA_STACK_CONTAINER_NAME=$(LLAMA_STACK_CONTAINER_NAME) \
	 LLAMA_STACK_IMAGE=$(LLAMA_STACK_IMAGE) \
	 FRESH=$(FRESH) \
	 bash scripts/manage_container.sh || exit 0
	@echo "Starting llama-stack container..."
	$(CONTAINER_RUNTIME) run -d \
		--name $(LLAMA_STACK_CONTAINER_NAME) \
		-p $(LLAMA_STACK_PORT):8321 \
```

anik120 (changes requested):
Thanks for the PR @Jazzcort. I agree with the principal of this - I initially put in "build every time" because that's best practice in general, so that existing dev environments always pick up the latest. However, in this case, we know lls version isn't going to change. So optimizing for faster rebuilds rather than fresh rebuilds makes sense to me 👍🏽 

Also need to run the linters/Black etc but I think you can see them already in CI.

tests/integration/container_lifecycle/test_container_lifecycle.py
Jazzcort: I agree! But since I changed `build-llama-stack-image` to be more destructive - remove existing image and rebuild it which will fail the test. The only way we can test the idempotency of building the image is to build it with a raw build command. I feel like if we want to test idempotancy then using raw commands for it is fine. We should separate the factor of how the Makefile behavior from it. We can create other set of tests for how `build-llama-stack-image` should behave. That's my thought. Let me know what do you think! Happy to discuss on this. 😁

Makefile
Jazzcort: I kind of leaning toward just leave it as it is. From my point of view, they're just 15 lines of code and they're not that complicated. The indirectness and those explicit variable forwarding make it not worth to extract into another script file. Also, I feel like this version should be the most complex one, I don't think we would add any more to the Makefile, at least for the part of image/container build process. Let me know what do you think! 😁

Jazzcort:
@anik120 Thanks for the insightful review! Let me know what do you think about my thoughts in the comments! 😁 We can work it out together!

tests/integration/container_lifecycle/test_container_lifecycle.py
Jazzcort: Just want to update something here.
If I understand it right, `tests/integration/container_lifecycle/test_container_lifecycle.py` is to test the behavior of Makefile acting as what we expect. Then, the reason why we test if the second image has the same id as the first one is because that's the behavior of the previous `build-llama-stack-image`. Since the current `build-llama-stack-image` has a different behavior, we should tailor the test cases for it. Therefore, I think we should just remove the assertion of `first_image_id == second_image_id` and rename the test `test_rebuild_after_clean_succeeds`. We just need to make sure the second build also succeed.

PR #1913, opened 2026-06-11
LCORE-1830: Implement Question Validity Safety Capability in Pydantic AI

Jazzcort:
## Description

Implement an LLM-based guardrail that classifies user questions as on-topic (Kubernetes/OpenShift or customized topic) before the main agent processes them. Off-topic questions are short-circuited with a rejection message, bypassing the primary agent entirely. Includes unit tests.

More thorough tests (e2e) can be implemented when we wire up the pydantic AI agent to our endpoints and see if this question validity shield really shields the agent from off-topic questions. 

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

The unit tests were drafted by Claude Code which was then reviewed and  refined by me. 😁 

[LCORE-1830](https://redhat.atlassian.net/browse/LCORE-1830)

## Testing
Since we have not wired pydantic_ai to our endpoints, you can only do an ad-hoc test with it.

## Summary by CodeRabbit

* **New Features**
  * Added a `QuestionValidity` capability to guard agent runs by classifying incoming prompts as allowed or rejected.
  * When rejected, execution is short-circuited and a configurable invalid-question response is returned.

* **Documentation**
  * Added module-level documentation for the capabilities packages and related unit test packages.

* **Tests**
  * Added unit tests covering prompt extraction/formatting, model settings sanitization, allowed/rejected control flow, and token-usage propagation.

### DIFF
--- a/src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
+++ b/src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
@@ -1,0 +1,151 @@
+"""Question validity capability for filtering off-topic user queries.
+
+This module implements a guardrail that classifies user questions as
+Kubernetes/OpenShift-related or not (It can be customized to any
+topic as well), using an LLM-based check before the main agent
+processes the request. Invalid questions are rejected with a
+predefined response, bypassing the primary agent entirely.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+from dataclasses import dataclass, field
+from string import Template
+
+from pydantic_ai import AgentRunResult, RunContext
+from pydantic_ai._agent_graph import GraphAgentState
+from pydantic_ai.capabilities import AbstractCapability, WrapRunHandler
+from pydantic_ai.direct import model_request
+from pydantic_ai.messages import ModelRequest, TextContent, UserContent
+from pydantic_ai.models import Model
+from pydantic_ai.models.openai import OpenAIResponsesModelSettings
+
+from client import AsyncLlamaStackClientHolder
+from log import get_logger
+from models.config import (
+    QuestionValidityConfig,
+)
+from pydantic_ai_lightspeed.llamastack import LlamaStackResponsesModel
+from utils.pydantic_ai import llama_stack_provider_from_client
+
+logger = get_logger(__name__)
+
+SUBJECT_REJECTED = "REJECTED"
+SUBJECT_ALLOWED = "ALLOWED"
+
+
+def _extract_message_str_from_user_content(user_content: Sequence[UserContent]) -> str:
+    """Extract and combine all text content into a string from a UserContent sequence.
+
+    Parameters:
+        user_content: A sequence of user content items to extract text from.
+
+    Returns:
+        A single string with all text content joined by newlines.
+    """
+    str_arr: list[str] = []
+    for c in user_content:
+        match c:
+            case str() as s:
+                str_arr.append(s)
+            case TextContent(content=c):
+                str_arr.append(c)
+
+    return "\n".join(str_arr)
+
+
+def _create_model_from_llama_stack_client(model_id: str) -> LlamaStackResponsesModel:
+    """Create a LlamaStackResponsesModel from the shared Llama Stack client.
+
+    Parameters:
+        model_id: The model identifier to use for the responses model.
+
+    Returns:
+        A configured LlamaStackResponsesModel instance.
+    """
+    client = AsyncLlamaStackClientHolder().get_client()
+    provider = llama_stack_provider_from_client(client)
+    settings = OpenAIResponsesModelSettings(openai_store=False)
+    return LlamaStackResponsesModel(model_id, provider=provider, settings=settings)
+
+
+@dataclass
+class QuestionValidity(AbstractCapability[None]):
+    """Block or modify user input based on a guardrail check.
+
+    The guard function receives the user prompt and returns True if safe.
+
+    Example:
+        ```python
+        from pydantic_ai import Agent
+        from pydantic_ai.models.openai import OpenAIResponsesModel
+
+        model = OpenAIResponsesModel("gpt-4o-mini")
+        agent = Agent("openai:gpt-4.1", capabilities=[QuestionValidity(model)])
+        ```
+    """
+
+    config: QuestionValidityConfig
+    _model: Model = field(init=False)
+
+    def __post_init__(self) -> None:
+        """Initialize the model instance from the configured model ID."""
+        self._model = _create_model_from_llama_stack_client(self.config.model_id)
+
+    def _build_prompt(self, message: str | Sequence[UserContent] | None) -> str:
+        """Build the classification prompt from the user message.
+
+        Parameters:
+            message: The user input as a string, sequence of user content, or None.
+
+        Returns:
+            The rendered prompt string ready to send to the validity model.
+        """
+        match message:
+            case str() as s:
+                _message = s
+            case Sequence() as seq:
+                _message = _extract_message_str_from_user_content(seq)
+            case None:
+                _message = ""
+
+        return Template(self.config.model_prompt).substitute(
+            message=_message, allowed=SUBJECT_ALLOWED, rejected=SUBJECT_REJECTED
+        )
+
+    async def wrap_run(
+        self, ctx: RunContext, *, handler: WrapRunHandler
+    ) -> AgentRunResult:
+        """Run the question validity check before delegating to the main agent.
+
+        Sends the user prompt to the validity model for classification.
+        If the question is allowed, the handler proceeds normally.
+        Otherwise, a rejection response is returned and the main agent
+        is bypassed.
+
+        Parameters:
+            ctx: The run context containing the user prompt and usage tracker.
+            handler: The handler that invokes the main agent run.
+
+        Returns:
+            The agent run result, either from the main agent or a rejection.
+        """
+        prompt = self._build_prompt(ctx.prompt)
+
+        result = await model_request(
+            model=self._model,
+            messages=[ModelRequest.user_text_prompt(prompt)],
+        )
+
+        # Include token usage from the question validity request
+        ctx.usage.incr(result.usage)
+
+        if result.text is not None and result.text.strip() == SUBJECT_ALLOWED:
+            return await handler()  # proceed with the real run
+
+        # short-circuit: return the rejection message with shield usage tracked
+        state = GraphAgentState(usage=ctx.usage)
+        return AgentRunResult(
+            output=self.config.invalid_question_response, _state=state
+        )
--- a/src/constants.py
+++ b/src/constants.py
@@ -250,6 +250,46 @@ DEFAULT_VIOLATION_MESSAGE: Final[str] = (
     "I cannot process this request due to policy restrictions."
 )
 
+# The Default model prompt and the default invalid question response for QuestionValidityConfig
+DEFAULT_MODEL_PROMPT: Final[str] = """
+Instructions:
+- You are a question classifying tool
+- You are an expert in kubernetes and openshift
+- Your job is to determine where or a user's question is related to kubernetes and/or openshift technologies and to provide a one-word response.
+- If a question appears to be related to kubernetes or openshift technologies, answer with the word ${allowed}, otherwise answer with the word ${rejected}.
+- Do not explain your answer, just provide the one-word response. Do not give any other response.
+- If the given question is an empty string, answer with the word ${rejected}
+
+
+Example Question:
+Why is the sky blue?
+Example Response:
+${rejected}
+
+Example Question:
+Why is the grass green?
+Example Response:
+${rejected}
+
+Example Question:
+Why is sand yellow?
+Example Response:
+${rejected}
+
+Example Question:
+Can you help configure my cluster to automatically scale?
+Example Response:
+${allowed}
+
+Question:
+${message}
+Response:
+"""
+DEFAULT_INVALID_QUESTION_RESPONSE: Final[str] = """
+Hi, I'm the OpenShift Lightspeed assistant, I can help you with questions about OpenShift, 
+please ask me a question related to OpenShift.
+"""
+
 # Placeholder slug used in responses when the server substituted its own
 # system prompt for the client's instructions.  Avoids leaking the actual
 # server prompt back to the client.
--- a/src/models/config.py
+++ b/src/models/config.py
@@ -2176,6 +2176,24 @@ class SkillsConfiguration(ConfigurationBase):
     )
 
 
+class QuestionValidityConfig(ConfigurationBase):
+    """Configuration for the question validity guardrail."""
+
+    model_id: str = Field(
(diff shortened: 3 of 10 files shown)


src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
```diff
@@ -0,0 +1,157 @@
...
+        match c:
+            case str() as s:
+                str_arr.append(s)
+            case TextContent(content=c):
+                str_arr.append(c)
+
+    return "\n".join(str_arr)
+
+
+def _remove_conversation_from_settings(model: Model) -> Model:
```
Jazzcort: During my testing, if we share `Model` either through passing the same Model instance as what is passed to Agent in `build_agent` or grabbing it from `ctx.model`, the request made by question validity shield contaminate the stored conversation history in Llama Stack. Therefore, the solutions that I can think of are either we copy the Model instance with "conversation" field removed from "extra_body" like this or we pass a separate Model instance without "conversation" field in "extra_body" when we create this capability in `build_agent`.

Jazzcort (commented):
The comment is just some bugs that I found when I tried to wire up this question validity shield with pydantic agent. 😁

jrobertboos (changes requested):
Instead of naming the files `capacity` they should be `capability`.

Jazzcort:
@jrobertboos  My bad! Thanks for catching that! 😁

Jazzcort:
Changed! @jrobertboos Thanks for reviewing!

src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
```diff
@@ -0,0 +1,163 @@
...
+        ```python
+        from pydantic_ai import Agent
+        from pydantic_ai.models.openai import OpenAIResponsesModel
+
+        model = OpenAIResponsesModel("gpt-4o-mini")
+        agent = Agent("openai:gpt-4.1", capabilities=[QuestionValidity(model)])
+        ```
+    """
+
+    model: Model
```
asimurka: Add a `config.py` module mirroring configuration from [providers](https://github.com/lightspeed-core/lightspeed-providers/blob/main/lightspeed_stack_providers/providers/inline/safety/lightspeed_question_validity/config.py)
Replace current attributes with config

src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
```diff
@@ -0,0 +1,163 @@
...
+                _settings = copy(settings)
+                _settings["extra_body"] = _extra_body
+                _model = copy(model)
+                _model._settings = _settings
+                return _model
+    return model
+
+
+@dataclass
+class QuestionValidity(AbstractCapability):
```
asimurka: Use `None` as type argument for `AbstractCapability` which corresponds to `Agent[None, str]` definition. Add type arguments corresponding to current lightspeed agent implementation where possible.

src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
```diff
@@ -0,0 +1,163 @@
...
+            case Sequence() as seq:
+                _message = _extract_message_str_from_user_content(seq)
+            case None:
+                _message = ""
+
+        return Template(self.model_prompt).substitute(
+            message=_message, allowed=SUBJECT_ALLOWED, rejected=SUBJECT_REJECTED
+        )
+
+    async def wrap_run(
```
asimurka, line 117: For discussion: overriding `before_model_request` method rather than `wrap_run` seems more natural for shields in my opinion.
For question validity, overriding `after_model_request` is not necessary.

src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
asimurka: Model used by question validity shield should be specified in lightspeed stack config not inherited from query request. Therefore, there should always be a separate model instance for the shield (see my other comments). Do you think that this will help with the interference?

asimurka (changes requested):

src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
Jazzcort: > or we pass a separate Model instance without "conversation" field in "extra_body" when we create this capability in build_agent.

If we're passing another model instance for this then we don't need this actually. I first came up with this because I don't know what will be passing when this gets created and I guess I was just trying to do some defensive programming there. 😆 Anyway, I will remove this and its unit test since we don't need it 😁

src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
Jazzcort: Yes! I'll do it! Thanks!

src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
Jazzcort: So I actually checked this from their codebase. 

```python
# TODO (v2): Change the default for all typevars like this from `None` to `object`
AgentDepsT = TypeVar('AgentDepsT', default=None, contravariant=True)
"""Type variable for agent dependencies."""

```

It seems like `None` is the default value if we don't pass it. But, we can still pass it to be more explicit. I'm fine with either ways. Let me know what do you think! 😁

src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
Jazzcort: So just one thing that I would like to bring up for consideration. The output from `before_model_request` if we bail out with `SkipModelRequest(message)` will still be processed by `after_model_request` and the Node lifecycle. Therefore, it has more chance to intertwine with other capabilities. 

Follow the agent run lifecycle 
run -> node_run -> model request -> ...
I feel like for question validity, we should have it as run-level since there is actually no need to further process the question downward if it's not valid.

It's just my thought. I'll follow whatever you guys think is the most suitable solution here. 😁

Jazzcort (commented):
Thanks @asimurka for the insightful review! I put my thoughts in the comments. Please let me know what do you think. 😁

src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
asimurka: If you look at the config I linked below you see that there is a `model_id`. So just use that and create separate model instance ;)

src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
asimurka: Yeah, that's true. It's up to you, just keep it consistent with redaction capability so it's not confusing.

src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
asimurka: Ok this sounds reasonable.

asimurka:
@Jazzcort when you're done with your changes please run `make verify` and fix all the linter errors

src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
Jazzcort: roger that!

asimurka:
@Jazzcort also please place configuration models into `src/models/config.py`. It will avoid circular dependencies and simplify documentation generation.

Jazzcort:
Got it! I move it there and also fix the Pydocstring check. The Pyright error is not coming from this PR so it should be fine.

Jazzcort:
@asimurka I think it should be good now. Let me know if there are any changes needed! 😁

src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
```diff
@@ -0,0 +1,132 @@
...
+        agent = Agent("openai:gpt-4.1", capabilities=[QuestionValidity(model)])
+        ```
+    """
+
+    config: QuestionValidityConfig
+    _model: Model = field(init=False)
+
+    def __post_init__(self) -> None:
+        """Initialize the model instance from the configured model ID."""
+        self._model = infer_model(self.config.model_id)
```
asimurka: This is conceptually wrong because we use Llama Stack as OpenRespones provider, not raw pydantic-ai inference, so you cannot use `infer_model`. You have to construct the model by yourself. For the reference see `utils/pydantic_ai.py`, namely `build_agent` function. 

Create llama stack provider from client (pass it as a private attribute f.e.), use minimal settings (store=false, no conversation etc.), create `LlamaStackResponsesModel` with previously created provider, settings and `model_name` reflecting `model_id` from config.

src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
```diff
@@ -0,0 +1,132 @@
...
+
+        Returns:
+            The agent run result, either from the main agent or a rejection.
+        """
+        prompt = self._build_prompt(ctx.prompt)
+
+        result = await model_request(
+            model=self._model,
+            messages=[ModelRequest.user_text_prompt(prompt)],
+        )
```
asimurka, line 139: If model is set up correctly, you can use its `request` method instead.

asimurka (changes requested):
I would like to hear @jrobertboos 's opinion on this but the fundamental problem is that our users are used to Llama Stack's  `provider/model` format and raw pydantic uses `provider:model` and I'm not sure if this is just about replacing `:` for `/`.

src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
Jazzcort: @asimurka  Just curious and want to know the reason why we can not use raw pydantic-ai inference here (why does it have to be coming from Llama Stack)? Do we have any constraints here? 😁

src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
Jazzcort: Model.request() feels like an internal function. model_request() is a thin wrapper that's more user facing I think. Also, if we're using request() then we force ourselves to pass `ModelRequestParameters` right? Just want some clarifications on this, is this what we want? 🤔

asimurka: Yeah I think it's ok to use model_request() I just saw some demo where they used the method but this is ok IMO.

src/pydantic_ai_lightspeed/capabilities/question_validity/_capability.py
asimurka: We have no constants but our consumers already use this shield with models supported by Llama Stack. I'm not sure what models they exactly use but those may not be supported by Pydantic, you never know. If we use Llama Stack provider, the migration will be much smoother for our customers.
Anyway this is just my opinion

Jazzcort: Got it! Thanks for filling me in! That makes sense! So we might have to address these model set in the future when we completely remove Llama Stack. I'll change it to use the OpenAIResponsesModel for now.

```python
def _create_model_from_llama_stack_client(model_id: str) -> OpenAIResponsesModel:
    client = AsyncLlamaStackClientHolder().get_client()
    provider = _llama_stack_provider_from_client(client)
    settings = OpenAIResponsesModelSettings(openai_store=False)
    return OpenAIResponsesModel(model_id, provider=provider, settings=settings)
```

And we create it in the `__post_init__`

```python
def __post_init__(self) -> None:
    self._model = _create_model_from_llama_stack_client(self.config.model_id)
```

Note that because the implementation will use `_llama_stack_provider_from_client()` from `src/utils/pydantic_ai.py`, we might need to remove the beginning "_" to indicate it's not a private helper function.

---
Does this looks good to you @asimurka @jrobertboos

jrobertboos: yes this is the best way :)

regarding:
> Note that because the implementation will use _llama_stack_provider_from_client() from src/utils/pydantic_ai.py, we might need to remove the beginning "_" to indicate it's not a private helper function.

I would just remove the "_" and use the methods from pydantic_ai.py and then import them into the capability. (increases PR scope a little bit but imo worth it)

asimurka: This looks good to me, but use `LlamaStackResponsesModel` from `_model.py` instead of raw `OpenAIResponsesModel`. This is just our thin wrapper around openai model. There should be no difference for shield usage now but you never know, Llama Stack is pretty creative with introducing new bugs and if so, it will be most likely fixed in this wrapper ;)

Jazzcort: Thanks for the review guys! I need a rebase to get `LlamaStackResponsesModel`. Will finish this up today! 😁

Jazzcort:
@asimurka @jrobertboos Just some concerns that I have for importing 'llama_stack_provider_from_client' from `utils/pydantic_ai.py` if we're planing to wire up the capabilities inside `build_agent` because we'll encounter a circular import issue if we do it.

Also I feel like `llama_stack_provider_from_client` and `_model_settings_from_responses_params` are more suitable to be placed at `src/pydantic_ai_lightspeed/llamastack/_provider.py` and `src/pydantic_ai_lightspeed/llamastack/_model.py` because they're both doing some customized process to prepare `LlamaStackProvider` and `LlamaStackResponsesModel`.

Check out this [branch](https://github.com/Jazzcort/lightspeed-stack/tree/restructure-pydantic-ai-utils-function) and let me know what you guys think! 😁

jrobertboos:
@Jazzcort I think that works, if you want to put up a PR with restructuring `pydantic_ai.py` into the `llamastack` package that would be great. Instead of putting it in `_provider.py` though I would add a `helper.py` file (or something like that) just to keep these helper/util functions isolated from the provider implementation. @asimurka do you have any objections to this?

asimurka:
That sounds reasonable. Is it gonna be included in this PR or in a separate one? Anyway current solution LGTM logically.

Jazzcort:
>Instead of putting it in _provider.py though I would add a helper.py file (or something like that) just to keep these helper/util functions isolated from the provider implementation. 

Sounds good to me!

@asimurka @jrobertboos  I think we can just merge this one as is. I'll do a separate PR for the restructure if both of you agree on the restructure idea. 😁

asimurka (approved):
LGTM, all the comments were addressed.

jrobertboos (approved):
LGTM

tisnik (approved):
LGTM

PR #1996, opened 2026-06-25
RSPEED-3313: inline single-use helpers and drop redundant None assignments in rlsapi_v1

major:
Pure refactor of `src/app/endpoints/rlsapi_v1.py` to reduce indirection and redundant code.

- Inline `_get_cla_version`, `_get_configured_default_model_name`, `_is_verbose_enabled`, and `_get_prompt_template` at their sole call sites
- Reduce `_compile_prompt_template` `lru_cache` maxsize from 8 to 1 (only one system prompt exists at a time)

No behavior change. All 63 existing unit tests pass without modification.


tisnik (approved):
LGTM

PR #2020, opened 2026-06-29
LCORE-2341-unified mode feature files

radofuchs:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [X] End to end tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Related Issue #LCORE-2341
- Closes #LCORE-2341

## Testing

## Summary by CodeRabbit

* **Tests**
  * Added new end-to-end coverage for unified-mode configuration across both library and server modes, including startup readiness checks and successful query execution.
  * Introduced additional E2E scenarios for unified-mode validation, legacy-to-unified config migration, and unified-mode synthesis (output location, file permissions, secret handling, and startup logging).
  * Included all new unified-mode feature specs in the main E2E test execution list.

### DIFF
--- a/tests/e2e/features/unified-mode-legacy.feature
+++ b/tests/e2e/features/unified-mode-legacy.feature
@@ -1,0 +1,39 @@
+@e2e_group_2
+Feature: Legacy two-file configuration during deprecation window
+
+  Background:
+    Given The service is started locally
+      And The system is in default state
+      And REST API service prefix is /v1
+      And the Lightspeed stack configuration directory is "tests/e2e/configuration"
+
+
+  # --- library mode (@skip-in-server-mode) ---
+
+  @skip-in-server-mode
+  Scenario: Legacy two-file configuration still boots and serves requests in library mode
+    Given The service uses the lightspeed-stack.yaml configuration
+      And The service is restarted
+     When I access endpoint "readiness" using HTTP GET method
+     Then The status code of the response is 200
+     When I use "query" to ask question
+     """
+     {"query": "Say hello", "model": "{MODEL}", "provider": "{PROVIDER}"}
+     """
+     Then The status code of the response is 200
+
+
+  # --- server mode (@skip-in-library-mode) ---
+
+  @skip-in-library-mode
+  Scenario: Legacy two-file configuration still boots and serves requests in server mode
+    Given The service uses the lightspeed-stack.yaml configuration
+      And Llama Stack is restarted
+      And Lightspeed Stack is restarted
+     When I access endpoint "readiness" using HTTP GET method
+     Then The status code of the response is 200
+     When I use "query" to ask question
+     """
+     {"query": "Say hello", "model": "{MODEL}", "provider": "{PROVIDER}"}
+     """
+     Then The status code of the response is 200
--- a/tests/e2e/features/unified-mode-legacy.feature
+++ b/tests/e2e/features/unified-mode-legacy.feature
@@ -1,4 +1,4 @@
-@e2e_group_2
+@e2e_group_2 @skip
 Feature: Legacy two-file configuration during deprecation window
 
   Background:
--- a/tests/e2e/features/unified-mode-migration.feature
+++ b/tests/e2e/features/unified-mode-migration.feature
@@ -1,4 +1,4 @@
-@e2e_group_2
+@e2e_group_2 @skip
 Feature: Legacy to unified configuration migration
 
   Background:
--- a/tests/e2e/features/unified-mode-migration.feature
+++ b/tests/e2e/features/unified-mode-migration.feature
@@ -17,15 +17,15 @@ Feature: Legacy to unified configuration migration
   Scenario: migrate then synthesize round-trips to the original run.yaml
      When lightspeed-stack --migrate-config is run for the legacy migration fixture pair
       And the active unified configuration is synthesized to run.yaml
-     Then the synthesized run.yaml is byte-identical to the legacy migration fixture run.yaml
+     Then the synthesized run.yaml parses to the same data as the legacy migration fixture run.yaml
 
 
   # --- library mode (@skip-in-server-mode) ---
 
   @skip-in-server-mode
   Scenario: Migrated unified configuration drives byte-identical Llama Stack behavior in library mode
-     When lightspeed-stack --migrate-config is run for the legacy migration fixture pair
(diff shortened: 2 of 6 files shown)


tisnik (approved):
LGTM. The When/Given steps switch looked weird at 1st glance, but seems to be ok.

tests/e2e/features/unified-mode-migration.feature
```diff
@@ -0,0 +1,53 @@
...
+  Scenario: migrate-config produces a unified configuration from a legacy pair
+     When lightspeed-stack --migrate-config is run for the legacy migration fixture pair
+     Then the file lightspeed-stack-unified-migrated.yaml contains native_override
+      And the file lightspeed-stack-unified-migrated.yaml does not contain library_client_config_path
+
+
+  Scenario: migrate then synthesize round-trips to the original run.yaml
+     When lightspeed-stack --migrate-config is run for the legacy migration fixture pair
+      And the active unified configuration is synthesized to run.yaml
+     Then the synthesized run.yaml is byte-identical to the legacy migration fixture run.yaml
```
max-svistunov: This asserts byte-identical, but the spec's round-trip test (see R4 in the design doc) is dict-equality, not byte-equality. Byte-identical YAML is fragile (things like key ordering, quote style, comment loss, anchor expansion all break it), and the synthesizer has no reason to preserve any of them.

How about:

```gherkin
Then the synthesized run.yaml parses to the same data as the legacy migration fixture run.yaml
```

(The scenario titles on L26/L42 saying "byte-identical behavior" are fine -- behavior is byte-identical per the spec, it's just the file itself that isn't.)

tests/e2e/features/unified-mode-legacy.feature
```diff
@@ -0,0 +1,39 @@
...
+    Given The service is started locally
+      And The system is in default state
+      And REST API service prefix is /v1
+      And the Lightspeed stack configuration directory is "tests/e2e/configuration"
+
+
+  # --- library mode (@skip-in-server-mode) ---
+
+  @skip-in-server-mode
+  Scenario: Legacy two-file configuration still boots and serves requests in library mode
```
max-svistunov, line 14: R2 has two halves -- `still boots and serves` AND `one startup deprecation WARN in 0.6; no WARN in unified mode`. This covers the boot half but not the WARN, and the WARN is really the point of R2 (it's how we prove the deprecation is actually live).

The log-assertion plumbing already exists -- synthesis.feature does `Then the lightspeed-stack container logs contain synthesized run.yaml`, so a `Then the lightspeed-stack container logs contain a legacy-config deprecation warning` here would mirror it (that exact string is just illustrative -- the real WARN wording is 2339's). Could you add a WARN-present scenario in legacy mode, and ideally a WARN-absent one in unified mode over in boot.feature?

tests/e2e/features/unified-mode-validation.feature
```diff
@@ -0,0 +1,19 @@
...
+  Scenario: inference.providers together with library_client_config_path fails at load
+    Given The service uses the lightspeed-stack-invalid-providers-and-legacy.yaml configuration
+     When configuration validation is attempted for the active configuration
+     Then the validation error contains --migrate-config
+
+
+  Scenario: llama_stack.config together with library_client_config_path fails at load
+    Given The service uses the lightspeed-stack-invalid-config-and-legacy.yaml configuration
+     When configuration validation is attempted for the active configuration
+     Then the validation error contains --migrate-config
```
max-svistunov, line 19: R11 in the spec is verified by both unit and e2e tests. The e2e part covers this case: a config declares a `config_format_version`, but that version does not match the config's actual shape, so loading must fail. This case is not tested anywhere yet -- the boot scenarios exercise shape auto-detection, but not the version mismatch.

Could you add one more scenario here, like the two above? For example, a config whose `config_format_version` says legacy while its body is unified, asserting that loading fails.

Lower priority than the deprecation-WARN gap; fine to move to a follow-up PR if you would rather keep this one small. I am flagging it because R11 lists e2e.

tests/e2e/features/unified-mode-migration.feature
```diff
@@ -0,0 +1,53 @@
...
+
+
+  # --- library mode (@skip-in-server-mode) ---
+
+  @skip-in-server-mode
+  Scenario: Migrated unified configuration drives byte-identical Llama Stack behavior in library mode
+     When lightspeed-stack --migrate-config is run for the legacy migration fixture pair
+    Given The service uses the lightspeed-stack-unified-migrated.yaml configuration
```
max-svistunov: Nit: `Given` after `When` reads backwards here -- the migrate is logically the precondition for "The service uses ...migrated.yaml". Behave parses it fine, so it is only cosmetic, but reversing the two reads more naturally:

```gherkin
Given lightspeed-stack --migrate-config has been run for the legacy migration fixture pair
And The service uses the lightspeed-stack-unified-migrated.yaml configuration
```

Same on L44.

max-svistunov (changes requested):
Nice work -- these track the spec's acceptance surface closely and reuse the existing steps well. A few things before merge: two I'd want addressed (the byte-identical assertion on the round-trip, and the missing deprecation-WARN coverage for R2), one optional (the R11 version-mismatch case), and a Gherkin nit.

Btw, reviewing surfaced a couple of step-implementation notes for 2343 -- captured on LCORE-2343, nothing blocking here.

tests/e2e/features/unified-mode-migration.feature
radofuchs: this was copy paste error, but completely agree that it reads weirdly. To keep the look unified across all scenarios, I will fix add the given step

When implementing, create this step as a step, this allows it to serve as both Given or When step

tests/e2e/features/unified-mode-migration.feature
radofuchs: agreed, will fix

max-svistunov (approved):
EDIT: Waiting on tests, then will re-approve.

max-svistunov (changes requested):
LGTM, thank you.

max-svistunov (approved):

PR #2146, opened 2026-07-14
LCORE-1427: Normalize Tool and Inline RAG default behavior

Jazzcort:
## Description
Remove fallback to all registered vector stores when tool RAG is unconfigured.

Previously, when neither rag.tool nor rag.inline were configured, prepare_tools would fetch and register all vector stores from llama-stack as a backward-compatibility fallback. This implicit behavior is removed — tool RAG now requires explicit configuration via rag.tool. Updated docs, config model, example YAML, and tests to reflect the new behavior.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Closes [LCORE-1427](https://redhat.atlassian.net/issues?filter=-1&selectedIssue=LCORE-1427)

## Testing
1. Set `rag.tool` to an empty list or simply omit it.
2. Check that there is no tool rag enabled even if there are some manually registered vector stores from llama-stack side.

## Summary by CodeRabbit

- **Behavior Changes**
  - RAG strategies are now disabled by default (`inline` and `tool` start as empty).
  - Tool-based RAG is only enabled when `rag.tool` (or per-request vector-store IDs) is explicitly provided.
  - If `rag.tool` is omitted, tool-based `file_search` RAG is disabled (no automatic inclusion of all BYOK stores).
  - Per-request vector-store IDs take precedence over `rag.tool`.

- **Documentation**
  - Updated configuration, API, BYOK guide, and examples to reflect explicit enablement rules.

- **Tests**
  - Refreshed unit and integration coverage for RAG enablement, vector-store resolution precedence, and tool selection behavior.
  - Added/updated container cleanup teardown for improved test hygiene.

### DIFF
--- a/src/utils/responses.py
+++ b/src/utils/responses.py
@@ -127,44 +127,6 @@ from utils.token_counter import TokenCounter
 logger = get_logger(__name__)
 
 
-async def get_vector_store_ids(
-    client: AsyncLlamaStackClient,
-    vector_store_ids: Optional[list[str]] = None,
-) -> list[str]:
-    """Get vector store IDs for querying.
-
-    If vector_store_ids are provided, returns them. Otherwise fetches all
-    available vector stores from Llama Stack.
-
-    Args:
-        client: The AsyncLlamaStackClient to use for fetching stores
-        vector_store_ids: Optional list of vector store IDs. If provided,
-            returns this list. If None, fetches all available vector stores.
-
-    Returns:
-        List of vector store IDs to query
-
-    Raises:
-        HTTPException: With ServiceUnavailableResponse if connection fails,
-            or InternalServerErrorResponse if API returns an error status
-    """
-    if vector_store_ids is not None:
-        return vector_store_ids
-
-    try:
-        vector_stores = await client.vector_stores.list()
-        return [vector_store.id for vector_store in vector_stores.data]
-    except APIConnectionError as e:
-        error_response = ServiceUnavailableResponse(
-            backend_name="Llama Stack",
-            cause=str(e),
-        )
-        raise HTTPException(**error_response.model_dump()) from e
-    except APIStatusError as e:
-        error_response = InternalServerErrorResponse.generic()
-        raise HTTPException(**error_response.model_dump()) from e
-
-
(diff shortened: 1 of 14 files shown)


Jazzcort:
The Ad-hoc free disk space step helps us to pass integration test more consistently, but not sure if it's a right thing to do. We maybe need to optimize `tests/integration/container_lifecycle/test_container_lifecycle.py` so it doesn't consume too much disk space with all those podman commands. Might need some clean up process after each test case. 😁

are-ces:
> The Ad-hoc free disk space step helps us to pass integration test more consistently, but not sure if it's a right thing to do. We maybe need to optimize `tests/integration/container_lifecycle/test_container_lifecycle.py` so it doesn't consume too much disk space with all those podman commands. Might need some clean up process after each test case. 😁

cc: @radofuchs

src/utils/responses.py
```diff
@@ -129,17 +128,11 @@
 
 async def get_vector_store_ids(
```
are-ces: As CodeRabbit points out this method is not used anymore, you can remove it 😄

are-ces (changes requested):
LGTM.
I had one nit to address, TY!

Jazzcort:
@are-ces @radofuchs Sorry, I forgot to update here. I took out the ad-hoc fix in the github workflow and added a teardown process that prunes all the dangling images after the container lifecycle test.

are-ces:
Since @radofuchs set that up I would like to have him review changes in `test_container_lifecycle.py`, they LGTM

src/utils/responses.py
Jazzcort: Will do!

are-ces (approved):
LGTM TY

tisnik (approved):
LGTM

PR #2234, opened 2026-07-23
LCORE-3202: Wire shield into response

Jazzcort:
## Description
Wire V2 shield moderation into the Responses API endpoint

Add run_shield_moderation_v2 and build_shield to utils/shields.py to run safety capabilities (question validity, PII redaction) from the new ShieldConfiguration model. Call the new function in the responses endpoint handler alongside the existing Llama Stack shield path.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/src/utils/shields.py
+++ b/src/utils/shields.py
@@ -3,15 +3,11 @@
 from typing import Optional
 
 from fastapi import HTTPException
-from ogx_client import (
-    APIConnectionError,
-    AsyncOgxClient,
-)
-from ogx_client import (
-    APIStatusError as LLSApiStatusError,
-)
+from ogx_client import APIConnectionError, APIStatusError, AsyncOgxClient
+from pydantic_ai.exceptions import AgentRunError
 
 from configuration import AppConfig
+from log import get_logger
 from models.api.requests import QueryRequest
 from models.api.responses.error import (
     InternalServerErrorResponse,
@@ -19,8 +15,21 @@ from models.api.responses.error import (
     ServiceUnavailableResponse,
     UnprocessableEntityResponse,
 )
-from models.common import ShieldModerationPassed, ShieldModerationResult
-from models.config import ShieldConfiguration
+from models.common.moderation import (
+    ShieldModerationPassed,
+    ShieldModerationResult,
+)
+from models.config import QuestionValidityConfig, RedactionConfig, ShieldConfiguration
+from pydantic_ai_lightspeed.capabilities.base import AbstractSafetyCapability
+from pydantic_ai_lightspeed.capabilities.question_validity._capability import (
+    QuestionValidity,
+)
+from pydantic_ai_lightspeed.capabilities.redaction._capability import (
+    PiiRedactionCapability,
+)
+from utils.agents.error_handler import map_agent_inference_error
+
+logger = get_logger(__name__)
 
 
 def validate_shield_ids_override(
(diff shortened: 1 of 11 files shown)


Jazzcort:
Should only check the last commit. This PR should land after those two previous commits.

This is just a draft PR. I removed refusal_message and moderation_id from the moderation result since I think we don't need those anymore. But then, we need to dig into all the functions that use moderation result and safely modify them so the missing fields won't cause errors. i.e. `OpenAIResponseObject` uses both moderation_id and refusal_message. Will it be safe to just remove id field and use moderation_result.message as output? Or, we will have a bigger refactor there for how moderation result is used. Just want to have more info before I move forward. 😁  @asimurka @jrobertboos

src/utils/shields.py
```diff
@@ -119,6 +129,35 @@ def validate_shield_ids_override(
...
+        shield = build_shield(shield_config)
+        shield_result = await shield.run(input_text)
+
+        if shield_result.decision == "blocked":
+            return shield_result
+
+    return ShieldModerationPassedV2()
+
+
+def build_shield(shield_config: ShieldConfiguration) -> AbstractSafetyCapability[Any]:
```
asimurka: use `object` as type var here as well

asimurka (commented):
LGTM, finally we can delete old `run_shield_moderation` and replace it with v2 but it's ok for now

src/utils/shields.py
```diff
@@ -60,6 +73,74 @@ def validate_shield_ids_override(
...
+        selected_shield_ids: Optional list of shield names to filter by.
+
+    Returns:
+        Result indicating if content was blocked or passed.
+
+    Raises:
+        HTTPException: 503 if the shield model is unreachable or returns an
+            HTTP error, 500 if the model returns an unexpected response.
+    """
+    selected_shield_configs = (
```
asimurka: Replace for `get_shields_for_request` function from `utils.shields` module - the function has more detailed verification steps.

src/utils/shields.py
```diff
@@ -60,6 +73,74 @@ def validate_shield_ids_override(
...
+    )
+
+    try:
+        for shield_config in selected_shield_configs:
+            shield = build_shield(shield_config)
+            shield_result = await shield.run(input_text)
+
+            if shield_result.decision == "blocked":
+                return shield_result
+    except (ModelHTTPError, ModelAPIError) as e:
```
asimurka: AFAIK OGX as a provider will propagate its own errors. Therefore, I recommend catching OGX's `ApiConnectionError` and raise 503, and also `ApiConnectionError` and raising corresponding erro response. See `map_agent_inference_error` function in `utils.agents.query.py` (I think you can reuse it here)

asimurka (changes requested):

src/utils/shields.py
Jazzcort: Will change! Thanks for pointing that out! 😁

asimurka: You can even extract that to the main handler and pass here already filtered shields... but that's really up to you ;)

src/utils/shields.py
Jazzcort: I was checking `_map_api_errors` function of the `OpenAIResponsesModel` in `pydantic_ai/models/openai.py` which transfers these OpenAI's errors (`APIConnectionError` and `APIStatusError`) into pydantic ai  errors (`ModelHTTPError` and `ModelAPIError`). But there might be some place where raw OpenAI errors can be raise.

Didn't know we have `map_agent_inference_error` which maps the errors pretty well, I'll use that instead. 😁

asimurka: Yeah, that's true but the problem is that we're not using raw openai provider but OGX (acting as openai provider). OGX does not raise native openai errors, so all the errors actually pass the `_map_api_errors` untouched.

src/utils/shields.py
Jazzcort: I'll keep it as is to make `run_shield_moderation_v2` more centralized, but no strong opinion here. 😁

src/utils/shields.py
Jazzcort: I spent too much time digging into this. The ogx exception that can be raised through `OgxServerTransport`/`OgxLibraryTransport` will be caught in `request()` of `AsyncAPIClient` at `openai/_base_client.py` and get coverted into openai's own `APIConnectionError`.

```python
            except Exception as err:
                log.debug("Encountered Exception", exc_info=True)

                if remaining_retries > 0:
                    await self._sleep_for_retry(
                        retries_taken=retries_taken,
                        max_retries=max_retries,
                        options=input_options,
                        response=None,
                    )
                    continue

                log.debug("Raising connection error")
                raise APIConnectionError(request=request) from err

```

which means `model_request()` and `agent.run()` won't raise ogx's `APIConnectionError` and `APIStatusError`.
These ogx exceptions will only be raised if we're directly using ogx client.

I'm like 80% positive on this, but let me know if I'm wrong.

After all, it does not matter that much. It's just that we can eliminate some exceptions from the try block if we're using the inference through pydantic ai. 😁

Jazzcort:
I moved `map_agent_inference_error` and `map_pydantic_agent_run_error` from `utils/agents/query.py` to `utils/agents/error_handler.py` because I was hitting a circular import error. Let me know if you guys think of a better solution. 😁

asimurka (approved):
LGTM

tisnik (approved):
LGTM

PR #2374, opened 2026-08-06
RHIDP-14083: extend customization profile for safety shield prompts

Jdubrick:
## Description

When a question_validity shield omits model_prompt or invalid_question_response, we now fill those at configuration load from the customization profile (then LCORE defaults). That way GET /v1/shields and the agent QV path see the effective text without copying the same prompts into YAML. This lets users include long model_prompts in the profile, rather than stuffing them inline in the yaml.

Resolution order:

1. Explicit value in lightspeed-stack.yaml always wins (including "")
2. If omitted / null: profile system_prompts.validation / query_responses.invalid_resp
3. If those profile keys are missing: LCORE defaults

Profile keys come from the profile Python module (PROFILE_CONFIG), not from YAML under customization:.

Example (omit the prompt fields and let the profile supply them):
```yaml
customization:
  profile_path: /path/to/profile.py
shields:
  - name: topic-guard
    provider_id: question_validity
    config:
      model_id: openai/gpt-4o-mini
      # model_prompt and invalid_question_response omitted on purpose
```
With a profile that looks like:
```python
PROFILE_CONFIG = {
    "system_prompts": {
        "validation": "Is this on-topic? ${message}",
        # ...
    },
    "query_responses": {
        "invalid_resp": "I can only answer questions about the product.",
    },
}
```
After load, both fields are set to the profile values. GET /v1/shields returns those resolved strings.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Assisted-by: Grok 4.5
- Generated by: Grok 4.5

- Related Issue # https://redhat.atlassian.net/browse/RHIDP-14083
- Closes # https://redhat.atlassian.net/browse/RHIDP-14083

## Testing

## Summary by CodeRabbit

- **New Features**
  - Question-validity shield prompts and responses can inherit values from the active profile or built-in defaults.
  - Explicit configuration values, including empty strings, take precedence over fallback values.
  - Effective validation and invalid-response settings are exposed through shield and profile responses.
  - Prompts support `${message}` substitution during agent execution; responses moderation evaluates raw input.

- **Documentation**
  - Updated configuration, API, shield guidance, examples, and response documentation to explain fallback behavior, precedence, empty values, and placeholders.

### DIFF
--- a/src/models/config.py
+++ b/src/models/config.py
@@ -3299,20 +3299,21 @@ class Configuration(ConfigurationBase):
             profile_invalid_resp = profile.get_invalid_resp()
 
         for shield in self.shields:
-            match shield.config:
-                case QuestionValidityConfig() as qv_config:
-                    if qv_config.model_prompt is None:
-                        qv_config.model_prompt = (
-                            profile_validation
-                            if profile_validation is not None
-                            else constants.DEFAULT_MODEL_PROMPT
-                        )
-                    if qv_config.invalid_question_response is None:
-                        qv_config.invalid_question_response = (
-                            profile_invalid_resp
-                            if profile_invalid_resp is not None
-                            else constants.DEFAULT_INVALID_QUESTION_RESPONSE
-                        )
+            if not isinstance(shield.config, QuestionValidityConfig):
+                continue
+            qv_config = shield.config
+            if qv_config.model_prompt is None:
+                qv_config.model_prompt = (
+                    profile_validation
+                    if profile_validation is not None
+                    else constants.DEFAULT_MODEL_PROMPT
+                )
+            if qv_config.invalid_question_response is None:
+                qv_config.invalid_question_response = (
+                    profile_invalid_resp
+                    if profile_invalid_resp is not None
+                    else constants.DEFAULT_INVALID_QUESTION_RESPONSE
+                )
         return self
 
     @model_validator(mode="after")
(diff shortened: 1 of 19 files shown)


src/models/config.py
```diff
@@ -3190,6 +3254,44 @@ def validate_shield_names_unique(self) -> Self:
...
+            # pylint: disable=no-member  # Pydantic nested model field
+            profile = self.customization.custom_profile
+        profile_validation = None
+        profile_invalid_resp = None
+        if profile is not None:
+            profile_validation = profile.get_validation()
+            profile_invalid_resp = profile.get_invalid_resp()
+
+        for shield in self.shields:
+            match shield.config:
```
asimurka: This could be simplified to `isinstance`

asimurka (commented):
LGTM in overall, I pointed out one nit.

src/models/config.py
Jdubrick: Thank you @asimurka, I've updated it to reflect using `isInstance`

Jdubrick:
@tisnik @asimurka am I able to get a re-review please?

asimurka (approved):

PR #2380, opened 2026-08-07
RSPEED-3444: fix REST API metrics middleware route discovery and root_path handling

### LINKED ISSUE #2379: [BUG] metrics are never recorded when root_path is configured
# Bug Description

## **Lightspeed Stack version**

`0.6.0rc2`

## **Describe the bug**

All REST API middleware-level metrics are silently never recorded when the service is deployed with a non-empty `root_path` (let's say `/api/lightspeed`).

`RestApiMetricsMiddleware` records metrics only for requests whose path matches a known application route. When `root_path` is set, a reverse proxy forwards requests with the full prefixed path (e.g. `/api/lightspeed/v1/infer`), but `app_routes_paths` contains only the unprefixed application paths (e.g. `/v1/infer`). The prefix must be stripped before the comparison.

The middleware has prefix-stripping code, but it reads `root_path` from the ASGI scope (`scope.get("root_path", "")`), which is always `""`. The actual configured value lives on the FastAPI application object (`app.root_path`), not in the scope. As a result, the stripping logic never activates, every request path fails the route check, and no middleware-level metrics are ever recorded.

## **To Reproduce**

1. Configure the service with a non-empty `root_path` (e.g. set `root_path: /api/lightspeed` in the service configuration).
2. Deploy behind any reverse proxy that forwards requests using the full prefixed path.
3. Send any request to a known endpoint (e.g. `/v1/infer`).
4. Scrape the `/metrics` endpoint: `ls_rest_api_calls_total` will have no samples, and `ls_response_duration_seconds` will have no observations.

## **Expected behavior**

All REST API middleware-level metrics should be recorded for every request to a known API endpoint, regardless of whether a path prefix is configured.

## **Screenshots or output**

- `increase(ls_rest_api_calls_total[120m])` → **No data**, even after confirmed successful requests.
- `increase(ls_llm_token_sent_total[120m])` → **Has data** for the same window.

So metrics recorded inside endpoint handlers (`ls_llm_token_sent_total` etc.) work correctly because they run regardless of the middleware path check. Only middleware-level metrics (`ls_rest_api_calls_total`, `ls_response_duration_seconds`) are affected.

## **Additional context**

The bug affects any deployment where `root_path` is configured. With an empty `root_path` (local development or deployments without a path prefix), paths match directly and metrics record correctly

The ASGI scope's `root_path` field is not the same as the FastAPI application's `root_path`. FastAPI stores the configured value on `app.root_path` for OpenAPI documentation; it does not write it into the ASGI scope on every request. The fix is to read `root_path` from `app.root_path` instead of the scope.

thepetk:
## Description

Fix REST API middleware-level metrics never recording. `FastAPI 0.137.0` ([fastapi#15745](https://github.com/fastapi/fastapi/pull/15745)) refactored `include_router()` to include routers with lazy `_IncludedRouter` wrapper objects in `app.routes` instead of copying individual `Route` instances. The [old list comprehension](https://github.com/thepetk/lightspeed-stack/blob/main/src/app/main.py#L295-L298) filtered with `isinstance(route, (Mount, Route, WebSocketRoute))`, which silently drops `_IncludedRouter` objects — leaving `app_routes_paths` containing only the 4 built-in FastAPI routes (`/openapi.json`, `/docs`, `/docs/oauth2-redirect`, `/redoc`). No application route ever matched, so metrics were never 
recorded regardless of the `root_path` setting.

FastAPI was pinned to 0.141.1 in the LCORE-2922 dependency update, merged via PR [#2358](https://github.com/thepetk/lightspeed-stack/pull/2358) / commit [`18c0876d`](https://github.com/thepetk/lightspeed-stack/commit/18c0876d).

Additionally, `RestApiMetricsMiddleware` read `root_path` from [`scope.get("root_path", "")`](https://github.com/thepetk/lightspeed-stack/blob/main/src/app/main.py#L215). In older versions of FastAPI/Starlette the scope field was never populated by the framework for this use case, so it always returned `""`, the prefix-stripping logic never activated, and every prefixed path failed to be recorded.

About the fix:

Replaced the route discovery list comprehension with FastAPI's official `iter_route_contexts()` API (PR [fastapi#15785](https://github.com/fastapi/fastapi/pull/15785), discussion [#15791](https://github.com/fastapi/fastapi/discussions/15791)), which correctly resolves all registered routes including the lazy `_IncludedRouter` wrappers introduced in 0.137. We also read `root_path` from `app.root_path` instead of the ASGI scope, which is the actual source of the configured value. Together these two changes ensure `app_routes_paths` is fully populated and that proxy-prefixed paths are correctly stripped, so middleware-level metrics record for every API endpoint regardless of deployment configuration.

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Assisted-by: Claude Sonnet 4.6
- Generated by: N/A

- Related to RSPEED-3444: Partially related
- Fixes https://github.com/lightspeed-core/lightspeed-stack/issues/2379

## Testing

**To reproduce:**

On `main`:

- Start the service locally with `service.root_path: /api/lightspeed` configured (that simulates 3scale), then send requests using the full prefixed path the proxy will forward:

```
GET /api/lightspeed/liveness
GET /api/lightspeed/readiness
```

- Check `/metrics` afterwards: `ls_rest_api_calls_total` should have zero samples despite successful responses. Metrics lke `ls_llm_token_sent_total` should be recorded cause they are being handled inside the scope of the endpoint.

**To verify the fix end-to-end:**

After the fix, repeat the same requests. `/metrics` should show:

```
ls_rest_api_calls_total{path="/liveness",status_code="200"} 2.0
ls_rest_api_calls_total{path="/readiness",status_code="200"} 1.0
```

**Unit test coverage:**

- `test_rest_api_metrics_strips_root_path` — updated to patch `app.root_path` directly (scope carries no `root_path`, matching actual runtime behaviour).
- `test_rest_api_metrics_no_root_path_unchanged` — unchanged; confirms empty `root_path` deployments are unaffected.
- `test_rest_api_metrics_uses_app_root_path_not_scope` — regression test; fails if the middleware is switched back to reading from the scope.
- `test_app_routes_paths_contains_application_routes` — new; asserts `app_routes_paths` contains application routes beyond the 4 FastAPI built-ins. Fails if the `iter_route_contexts()` call is reverted to the old `isinstance` filter.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved API metrics path handling when an application is configured with a root path.
  - Metrics now correctly match routes and record request paths, including versioned endpoints such as `/v1/infer`.
  - Resolved inconsistencies when incoming requests do not include root-path information.

### DIFF
--- a/src/app/main.py
+++ b/src/app/main.py
@@ -10,7 +10,8 @@ from fastapi import FastAPI, HTTPException
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
 from ogx_client import APIConnectionError, AsyncOgxClient
-from starlette.routing import Mount, Route, WebSocketRoute
+from fastapi.routing import iter_route_contexts
+
 from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
 import version
@@ -212,7 +213,7 @@ class RestApiMetricsMiddleware:  # pylint: disable=too-few-public-methods
         # requests with the full prefixed path (/api/lightspeed/v1/infer) but
         # app_routes_paths contains only application-level paths (/v1/infer).
         # Strip the prefix so the path check and metric labels match the routes.
-        root_path = scope.get("root_path", "")
+        root_path: str = app.root_path
         path: str = scope["path"]
         if root_path and path.startswith(root_path + "/"):
             path = path[len(root_path) :]
@@ -292,9 +293,9 @@ logger.info("Including routers")
 routers.include_routers(app)
 
 app_routes_paths = [
-    route.path
-    for route in app.routes
-    if isinstance(route, (Mount, Route, WebSocketRoute))
+    rc.original_route.path
+    for rc in iter_route_contexts(app.routes)
+    if hasattr(rc.original_route, "path") and rc.original_route.path
 ]
 
 # Register pure ASGI middlewares.  Middleware execution order is the reverse of
--- a/tests/unit/app/test_main_middleware.py
+++ b/tests/unit/app/test_main_middleware.py
@@ -9,7 +9,14 @@ from fastapi import HTTPException, status
 from pytest_mock import MockerFixture
 from starlette.types import Message, Receive, Scope, Send
 
-from app.main import GlobalExceptionMiddleware, RestApiMetricsMiddleware
+from app.main import (
+    GlobalExceptionMiddleware,
+    RestApiMetricsMiddleware,
+    app_routes_paths,
+)
+from app.main import (
+    app as fastapi_app,
+)
 from models.api.responses.error import InternalServerErrorResponse
 
 
@@ -189,6 +196,7 @@ async def test_rest_api_metrics_strips_root_path(
 ) -> None:
     """Middleware must strip root_path so prefixed requests still match routes."""
     mocker.patch("app.main.app_routes_paths", ["/v1/infer"])
+    mocker.patch.object(fastapi_app, "root_path", "/api/lightspeed")
     mock_measure_duration = mocker.patch(
         "app.main.recording.measure_response_duration", return_value=nullcontext()
     )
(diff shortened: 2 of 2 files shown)


thepetk:
Will mark it again as "ready for review" once I complete a final round of tests

thepetk:
I see the same failures in many different PRs currently opened. I don't believe they are related to my changes tbh.

thepetk:
> Will mark it again as "ready for review" once I complete a final round of tests

PR updated. Was able to reproduce and verify the fix works

tisnik (approved):
LGTM

PR #2319, opened 2026-08-03
LCORE-2338: LS container entrypoint + deployment artifacts for unified mode

max-svistunov:
## Description

Implements LCORE-2338: server mode works end to end from a single unified `lightspeed-stack.yaml`.

The ticket assumed the Python CLI already auto-detected unified vs legacy configs — it did not: `main()` only performed legacy enrichment, so a container handed a unified config (no external run.yaml) could not start. This PR closes that gap and updates the deployment artifacts around it:

- **CLI auto-detection** (`src/llama_stack_configuration.py`): `main()` dispatches on a new `has_synthesis_input()` helper mirroring the root `Configuration.check_unified_vs_legacy` detection (non-empty `inference.providers`, non-empty `vector_store.providers`, or a `llama_stack.config` block). Unified configs are synthesized via `synthesize_to_file` (`--input` ignored and need not exist; relative `profile:` resolves against the `--config` dir, R8; output keeps mode 0600, R10). Legacy configs enrich `--input` exactly as before, so existing layouts and the CI provider-matrix workflows are untouched.
- **Entrypoint** (`scripts/llama-stack-entrypoint.sh`): documents the two startup modes, adopts the post-rename `ogx stack run` command, and fails with a clear error when generation fails and no fallback run.yaml is mounted. (Note for a follow-up: `ogx stack run` already emits a FutureWarning recommending `ogx run` — inherited from the rename, kept for consistency with the rest of main.)
- **Container image** (`deploy/llama-stack/test.containerfile`): ships `src/data/` to `/opt/app-root/data` so `load_default_baseline()` resolves next to the standalone-copied script.
- **Compose** (`docker-compose.yaml`): one file serves both modes — mode is chosen by the *content* of `lightspeed-stack.yaml`; the `run.yaml` mount is documented as legacy-only and is inert in unified mode. Adds the `src/data` host-copy mount beside the existing script mounts.
- **Docs** (`deploy/llama-stack/README.md`, new): image contents, the two startup modes, a minimal unified-only compose snippet (no run.yaml mount), and rebuild guidance.

`.tekton/` pipelines reference no compose files and need no changes (flagging for @radofuchs as the Konflux owner per the ticket).

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] Bump-up dependent library
[...]

PR #2401, opened 2026-08-10
LCORE-3520 Close two synthesis-mode gaps: LLM model registration and CLI access

omkarjoshi0304:
## Summary

While adopting unified synthesis mode (LCORE-2336) in the OpenStack Lightspeed operator, we ran into two gaps:

1. **LLM models aren't registered in `registered_resources.models`.** `apply_high_level_inference()` builds `providers.inference` entries from `inference.providers`, but never registers those models as resources. Without that registration, Llama Stack can only discover a model via `list_models()` at startup, which needs a live connection to the provider — so inference breaks if the endpoint is briefly unreachable during startup. We were working around this with a post-processing step in our own init container script.

2. **The CLI only supports legacy enrichment mode.** `python llama_stack_configuration.py -c config.yaml` calls `generate_configuration()`, which expects an already-built `run.yaml`. There's no CLI path to `synthesize_to_file()`/`synthesize_configuration()` (unified mode), so we had to import the module directly instead of using the documented script interface.

## Changes

- `apply_high_level_inference()` now registers each provider's `allowed_models` as an LLM resource in `registered_resources.models`, deduped against anything already registered.
- Added a `--synthesize` flag to the CLI: when set, builds a complete config via `synthesize_to_file()` from `-c` alone (ignoring `-i`), instead of enriching an existing `run.yaml`.

Both changes are additive — default CLI behavior for existing legacy-mode consumers is unchanged.

Once merged, we'll be able to call the plain CLI directly (`python -m llama_stack_configuration -c lightspeed-stack.yaml -o run.yaml --synthesize`) from our init container and drop our custom wrapper script entirely.

## Test plan

- [x] `uv run python -m pytest tests/unit/test_llama_stack_synthesize.py tests/unit/test_llama_stack_configuration.py -q` — 112 passed
- [x] `uv run ruff check` — clean
- [x] `uv run mypy src/llama_stack_configuration.py` — clean
- [x] `uv run black --check --fast` — clean
- [x] Rebased cleanly onto latest `upstream/main`

## Summary by CodeRabbit

* **New Features**
  * Added unified configuration synthesis through the `--synthesize` command-line option.
  * Automatically registers configured non-embedding models for high-level inference.
  * Supports multiple allowed models while preventing duplicate registrations.
  * Resolves relative configuration paths and handles empty configurations.
  * Preserves existing behavior when synthesis is not enabled.

* **Bug Fixes**
  * Prevented embedding models and outdated provider entries from being registered.
  * Added validation to reject duplicate provider identifiers and incompatible inputs.
  * Improved handling for providers without specified models.
  * Updated command-line descriptions and legacy input guidance.

### DIFF
--- a/src/llama_stack_configuration.py
+++ b/src/llama_stack_configuration.py
@@ -1064,6 +1064,130 @@ def _strip_default_openai_inference(ls_config: dict[str, Any]) -> None:
     ]
 
 
+def _replace_or_append_inference_provider(
+    inference_list: list[Any], entry: dict[str, Any]
+) -> None:
+    """Replace an inference entry with the same provider_id, else append.
+
+    Baseline ${env.OPENAI_API_KEY:+openai} matches as "openai" (LCORE-3607).
+
+    Parameters:
+        inference_list: Mutable providers.inference list.
+        entry: New provider entry to install.
+    """
+    provider_id = entry["provider_id"]
+    for index, existing in enumerate(inference_list):
+        if not isinstance(existing, dict):
+            continue
+        existing_id = _matchable_provider_id(existing.get("provider_id"))
+        if existing_id == provider_id:
+            logger.info(
+                "Replacing existing inference provider with "
+                "provider_id=%r; a later high-level entry overwrote it",
+                provider_id,
+            )
+            inference_list[index] = entry
+            return
+    inference_list.append(entry)
+
+
+def _build_inference_entry(
+    provider: dict[str, Any], emitted_id: str, ls_provider_type: str
+) -> tuple[dict[str, Any], list[str]]:
+    """Build a providers.inference entry from one high-level provider.
+
+    Parameters:
+        provider: One high-level ``inference.providers`` entry.
+        emitted_id: The provider_id to emit (explicit id or hyphenated type).
+        ls_provider_type: Llama Stack provider_type from :data:`PROVIDER_TYPE_MAP`.
+
+    Returns:
+        tuple[dict[str, Any], list[str]]: The provider entry, and its
+        ``allowed_models`` (empty list when unset).
+    """
+    entry: dict[str, Any] = {
+        "provider_id": emitted_id,
+        "provider_type": ls_provider_type,
+    }
+
+    provider_config: dict[str, Any] = {}
+    if provider.get("extra"):
+        provider_config.update(provider["extra"])
+    if provider.get("api_key_env"):
+        key_field = API_KEY_FIELD_MAP.get(ls_provider_type, "api_key")
+        provider_config[key_field] = "${env." + provider["api_key_env"] + "}"
+    allowed_models = provider.get("allowed_models") or []
+    if allowed_models:
+        provider_config["allowed_models"] = allowed_models
+    if provider_config:
+        entry["config"] = provider_config
+
+    return entry, allowed_models
+
+
+class _LLMModelRegistrar:  # pylint: disable=too-few-public-methods
+    """Owns ``registered_resources.models`` writes for one synthesis call.
+
+    ``apply_high_level_inference`` needs to register a model per
+    ``allowed_models`` entry, dedupe against models registered before the
+    call (baseline, native_override, BYOK, ...), and — when a later
+    high-level entry reuses a ``provider_id`` — evict the models it
+    registered for that provider's earlier declaration. Bundling those three
+    pieces of state here keeps that bookkeeping out of the caller instead of
+    threading a list, a set, and a dict through free-function parameters.
+    """
+
+    def __init__(self, ls_config: dict[str, Any]) -> None:
+        self._models = ls_config.setdefault("registered_resources", {}).setdefault(
+            "models", []
+        )
+        self._known_ids = {
+            m.get("model_id") for m in self._models if isinstance(m, dict)
+        }
+        self._owned_by_provider: dict[str, list[str]] = {}
+
+    def sync(self, provider_id: str, allowed_models: list[str]) -> None:
+        """Register ``allowed_models`` for ``provider_id``, replacing its prior set.
+
+        A later high-level entry with the same emitted ``provider_id`` fully
+        replaces the earlier one's provider config (see
+        ``_replace_or_append_inference_provider``), so any model this
+        registrar added for it earlier in the same call is stale and must be
+        evicted first — otherwise a model no longer served by the replaced
+        provider would linger in ``registered_resources.models``.
+
+        Parameters:
+            provider_id: provider_id of the inference provider offering the
+                models.
+            allowed_models: Model names to register.
+        """
+        stale = set(self._owned_by_provider.pop(provider_id, []))
+        if stale:
+            self._models[:] = [
+                m
+                for m in self._models
+                if not (isinstance(m, dict) and m.get("model_id") in stale)
+            ]
+            self._known_ids.difference_update(stale)
+
+        added = []
+        for model_name in allowed_models:
+            if model_name in self._known_ids:
+                continue
+            self._models.append(
+                {
+                    "model_id": model_name,
+                    "model_type": "llm",
+                    "provider_id": provider_id,
+                    "provider_model_id": model_name,
+                }
+            )
+            self._known_ids.add(model_name)
+            added.append(model_name)
+        if added:
+            self._owned_by_provider[provider_id] = added
+
+
 def apply_high_level_inference(
     ls_config: dict[str, Any], inference: dict[str, Any]
 ) -> None:
@@ -1082,6 +1206,14 @@ def apply_high_level_inference(
     Secrets are emitted as ``${env.<VAR>}`` references, never resolved
     values (R6).
 
+    Each of the provider's ``allowed_models`` is also registered as an LLM
+    entry in ``registered_resources.models`` (skipping any ``model_id`` already
+    present), so the model is usable even when the provider endpoint is
+    unreachable at startup — Llama Stack's auto-discovery otherwise requires a
+    live connection to list models. Replacing a provider (same emitted id)
+    also evicts the LLM entries this function registered for the earlier
+    declaration, so a superseded provider's models don't linger.
+
     Parameters:
         ls_config: The OGX configuration being synthesized (modified in
             place).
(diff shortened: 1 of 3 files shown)


max-svistunov (commented):
*NOTE: The review is written by Claude, I confirm it is correct.*

Thanks for this — both gaps are real, and the missing model registration is a genuine hole in what LCORE-2336 shipped.

The CLI half of the change conflicts with work already in review. PR #2319 (LCORE-2338) adds unified-mode dispatch to `main()` using auto-detection rather than an explicit flag: `has_synthesis_input(config)`, which reads the same synthesis inputs as the root `Configuration` model. That approach is specified in `docs/design/llama-stack-config-merge/llama-stack-config-merge.md` under "Trigger mechanism" — "The Python CLI auto-detects unified vs legacy by the same synthesis-input check."

Two other in-review PRs already depend on the CLI being flagless:

* #2448 (LCORE-2343) invokes it from the behave suite as `["src/llama_stack_configuration.py", "-c", source, "-o", output]`, with a step docstring stating it "runs the config CLI exactly as the server entrypoint does". Under flag-only dispatch that step falls through to legacy enrichment and the migrate-then-synthesize round-trip assertion fails.
* #2319 also rewrites `scripts/llama-stack-entrypoint.sh` around the detection. With a flag, the entrypoint would need to decide when to pass it, which means reimplementing the detection logic in bash.
* #2450 (LCORE-2345) rewrites the documentation to make unified mode primary, so a new user-facing flag would need to be documented there as well.

Would you consider dropping the second commit and letting #2319 provide the CLI path? Your init container still gets to drop its wrapper script; it just calls the CLI without a flag. If you need an explicit override — for example, to fail loudly rather than fall back to legacy on a malformed config — I'd suggest `--mode {auto,unified,legacy}` layered on top of the detection rather than a boolean that bypasses it.

Two process notes:

* The title prefix `OSPRH:32718` is not in the allowed list in `.github/pr-title-checker-config.json`, which is why the `check` job is red. It needs a ticket key from that list — an LCORE ticket under the unified-config epic (LCORE-836) would be the natural home.
* The repository is currently under a merge freeze with no announced end date, so neither this PR nor #2319 will land immediately. That removes any urgency about which lands first.

On the registration itself: I verified the premise rather than assuming it. The OpenAI-mixin providers short-circuit `register_model()` when `model_validation` is falsy (`ogx/providers/utils/inference/openai_mixin.py:535-548`), so a pre-registered model does not require a live endpoint. Llama Stack builds the identifier as `provider_id/model_id` (`ogx/core/routing_tables/models.py:376-380`), which matches what auto-discovery would produce, so existing references such as `openai/gpt-4o-mini` are unaffected. The approach is sound.

One addition to the test plan: `tests/integration/test_unified_synthesis.py` is the R7 parity suite covering this function. I ran it against your branch and it passes (15 passed, 1 xfailed), but it should be listed alongside the unit tests.

Specific comments inline.

src/llama_stack_configuration.py
```diff
@@ -998,6 +998,125 @@ def deep_merge_list_replace(
...
+            self._models[:] = [
+                m
+                for m in self._models
+                if not (isinstance(m, dict) and m.get("model_id") in stale)
+            ]
+            self._known_ids.difference_update(stale)
+
+        added = []
+        for model_name in allowed_models:
+            if model_name in self._known_ids:
```
max-svistunov: The deduplication key is the bare `model_id`, but Llama Stack scopes model identifiers by provider — `register_model` builds `f"{provider_id}/{model_id}"` unless the entry is flagged as an unprefixed alias (`ogx/core/routing_tables/models.py:367-380`). Two providers offering the same model name therefore do not collide in Llama Stack, and this check silently discards the second registration:

    providers: [{type: openai, id: openai, allowed_models: [gpt-4o]},
                {type: azure,  id: azure,  allowed_models: [gpt-4o]}]

    registered_resources.models -> [{"model_id": "gpt-4o", "provider_id": "openai", ...}]
    providers.inference          -> ["openai", "azure"]

`azure/gpt-4o` is dropped even though the provider is present in `providers.inference`. This does not affect a single-provider deployment, but serving the same model from two endpoints is a common configuration.

Consider keying on `(provider_id, model_id)` instead, including the `_known_ids` seeding in `__init__` — a baseline entry for `gpt-4o` under a different provider should not suppress this one. Either way, it would help to log the skip; a bare `continue` makes "my model was not registered and nothing explained why" hard to diagnose.

src/llama_stack_configuration.py
```diff
@@ -998,6 +998,125 @@ def deep_merge_list_replace(
...
+                continue
+            self._models.append(
+                {
+                    "model_id": model_name,
+                    "model_type": "llm",
```
max-svistunov: Minor: `sentence_transformers` is a valid `UnifiedInferenceProvider.type` and is an embedding provider, so `allowed_models` on it would register embedding models as `llm` and route them incorrectly. Consider skipping provider types that are not LLM providers, or adding a comment noting that `allowed_models` on the inline embedder is unsupported.

src/llama_stack_configuration.py
```diff
@@ -998,6 +998,125 @@ def deep_merge_list_replace(
...
+    allowed_models = provider.get("allowed_models") or []
+    if allowed_models:
+        provider_config["allowed_models"] = allowed_models
+    if provider_config:
+        entry["config"] = provider_config
+
+    return entry, allowed_models
+
+
+class _LLMModelRegistrar:  # pylint: disable=too-few-public-methods
```
max-svistunov: Is the eviction logic worth its complexity? It only triggers when two entries in `inference.providers` emit the same `provider_id`, which is a configuration error rather than a case worth resolving silently as last-wins. Since `UnifiedInferenceProvider` already has an `id` field, rejecting the duplicate during validation seems cleaner, and would let this collapse into a plain function returning a new list:

    def register_high_level_models(models, provider_id, allowed_models) -> list[dict]

rather than a class that holds an alias into `ls_config` and rewrites it through `self._models[:] = ...`. The project's CLAUDE.md favours returning new structures over in-place mutation, and `# pylint: disable=too-few-public-methods` on a class with one public method usually indicates a function.

If you prefer to keep the eviction, that's a reasonable call — but it would be worth filing the duplicate-id validation as a follow-up so the behaviour is not relied upon indefinitely.

src/llama_stack_configuration.py
```diff
@@ -1029,39 +1156,18 @@ def apply_high_level_inference(
 
     providers_section = ls_config.setdefault("providers", {})
     inference_list = providers_section.setdefault("inference", [])
+    model_registrar = _LLMModelRegistrar(ls_config)
```
max-svistunov: Registration happens before the `native_override` merge, and `registered_resources.models` is a list, so R5's list-replacement semantics discard the registrations wholesale when an operator's override supplies its own models. Verified:

    native_override.registered_resources.models = [{model_id: from-override, provider_id: openai}]
    result -> [('from-override', 'openai')]     # gpt-4o-mini removed, no warning

This is defensible under R5, which documents `native_override` as taking precedence over the high-level sections. It is worth being deliberate about it, though, because #2449 (LCORE-3370) applies the mirror-image fix for enrichment: enrichment moves after the override precisely because a migrated config — whose `native_override` is a lifted `run.yaml` — was silently discarding BYOK and Solr registrations. Once #2449 lands, BYOK's embedding models will be protected from this and the LLM models registered here will not, leaving two classes of auto-registered model with opposite precedence.

My suggestion is to keep the current behaviour for consistency with R5 and add a line to the `apply_high_level_inference` docstring noting that the registration can be overridden.

Separately, a minor point: constructing the registrar calls `setdefault` on `registered_resources.models`, so any config with `inference.providers` now produces `registered_resources: {models: []}` even when nothing is registered (asserted by `test_apply_high_level_inference_no_allowed_models_no_registration`). That is harmless to Llama Stack but changes the synthesized output for every existing unified config. Deferring the `setdefault` to the first actual append would avoid it.

src/llama_stack_configuration.py
```diff
@@ -1387,20 +1493,32 @@ def main() -> None:
...
+        "--synthesize",
+        action="store_true",
+        help="Build a complete run.yaml from -c alone instead of enriching "
+        "an existing run.yaml given by -i",
     )
     args = parser.parse_args()
 
     with open(args.config, "r", encoding="utf-8") as f:
-        config = yaml.safe_load(f)
+        config = yaml.safe_load(f) or {}
```
max-svistunov, line 1563: This change is not limited to synthesize mode. Legacy mode previously raised on `config.get(...)` with an empty `-c` file and now quietly copies the input through. That is arguably the better behaviour, but the effect is broader than the commit message describes.

src/llama_stack_configuration.py
```diff
@@ -1387,20 +1493,32 @@ def main() -> None:
...
     )
     parser.add_argument(
         "-o",
         "--output",
         default="run_.yaml",
-        help="Output enriched config (default: run_.yaml)",
+        help="Output config file (default: run_.yaml)",
+    )
+    parser.add_argument(
+        "--synthesize",
```
max-svistunov, line 1547: If the flag is retained: `-i` is silently ignored when `--synthesize` is set. Consider `default=None` on `-i` plus a `parser.error` when both are passed explicitly, so an argument the user supplied is not discarded without notice.

src/llama_stack_configuration.py
```diff
@@ -1387,20 +1493,32 @@ def main() -> None:
...
     )
     args = parser.parse_args()
 
     with open(args.config, "r", encoding="utf-8") as f:
-        config = yaml.safe_load(f)
+        config = yaml.safe_load(f) or {}
 
-    generate_configuration(args.input, args.output, config)
+    if args.synthesize:
+        synthesize_to_file(
```
max-svistunov, line 1566: This is a heads-up rather than a change request. `synthesize_to_file` chmods its output to 0600 (spec requirement R10), whereas the legacy path writes at the process umask. Measured locally: 600 for the synthesized file, 664 for the legacy one. If Llama Stack runs under a different uid than the init container performing the synthesis, it will not be able to read `run.yaml`. This came up during the e2e work, so it is worth checking against your operator's pod spec before switching over.

src/llama_stack_configuration.py
omkarjoshi0304: dedup now keyed on (`provider_id, model_id`) so the same model served by two  providers registers once per provider. Added `logger.debug` on skip.

src/llama_stack_configuration.py
omkarjoshi0304: added `EMBEDDING_PROVIDER_TYPES` constant; `sentence_transformers` is skipped for LLM registration with a debug log.

src/llama_stack_configuration.py
omkarjoshi0304: Added `check_unique_provider_ids` validator to `InferenceConfiguration`; duplicates are now rejected at load time. Replaced `_LLMModelRegistrar` with a plain `_register_high_level_models()` function.

src/llama_stack_configuration.py
omkarjoshi0304: ` registered_resources.models` is now created lazily (only when something actually registers). Added a docstring note that `native_override` (R5 list-replacement) takes precedence over these registrations.

src/llama_stack_configuration.py
omkarjoshi0304: added a comment noting the guard applies to both modes, not just synthesize.

src/llama_stack_configuration.py
omkarjoshi0304: `-i` defaults to `None`; passing both now calls `parser.error()`.

PR #2427, opened 2026-08-13
LCORE-1574: Integration tests for conversation compaction

Jazzcort:
## Description

Introduce an InMemoryConversationStore in conftest.py to provide stateful conversation item persistence during tests, and extend AsyncOgxClientHolder patches to cover a2a, responses, streaming_query, and utils.endpoints modules.

Add comprehensive integration test suite covering compaction triggering, existing marker reuse, small-conversation bypass, disabled-compaction passthrough, additive multi-round summarization, concurrent request blocking, and (for streaming) SSE event ordering and start-event deduplication.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [x] Integration tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Related Issue #
- Closes [LCORE-1574](https://redhat.atlassian.net/issues?filter=-1&selectedIssue=LCORE-1574)

## Testing
Run the integration test and all the newly added tests should all pass.

## Summary by CodeRabbit

- **Tests**
  - Added an in-memory conversation-item store for integration tests.
  - Conversation items now persist within individual tests and support creation, listing, and typed message conversion.
  - Expanded mock client coverage for conversation items, streaming queries, responses, utility endpoints, and A2A interactions.

- **Documentation**
  - Documented the conversation-store fixture, including setup, pre-population, inspection, and direct usage examples.


Jazzcort:
Hey @max-svistunov, here is the huge test suite for the compaction feature which covers all the endpoints that use it. Since the feature is implemented by you, you're the most suitable person to review this I think. 😁 Let me know if there are any changes needed.

Jazzcort:
Integration tests failed for `container_lifecycle`. That part of the integration test seems to be quite problematic these days.

Jazzcort:
Rebase to main which bypasses container lifecycle test and makes integration test green again 😁 

Please ignore the typos in my branch name. 😆

max-svistunov (commented):
@Jazzcort Nice work, the coverage is broad. 👍 But we need to increase coverage by one layer in two places:

* The compacted-mode assertions end up with `build_agent` args as opposed to with what reaches the model. Then we immediately get `prompt = cast(str, responses_params.input)` (`utils/agents/query.py:308` and the same at `streaming.py:395`). Which is a list in compacted mode, and `agent.run()` doesn't accept it => we get error 500 on both endpoints (my fix for this is currently on review #2451). This is accepted by mocked agent without warning. So we should assert on `mock_query_agent.run.call_args` (and the same for streaming).

* There is no conversation_cache in `tests/configuration/lightspeed-stack.yaml`, so everything here is marker-only => recursive folding doesn't get executed. I suggest `conversation_cache: {type: memory}` and one fold test.

The rest of the comments are written by Claude, they are less major, I checked them, they're valid.

tests/integration/conftest.py
```diff
@@ -441,6 +446,86 @@ def set_streaming_query_agent_run(
...
+# ==========================================
+
+
+class InMemoryConversationStore:
+    """In-memory fake for the Llama Stack conversations.items API.
+
+    Provides stateful ``list`` and ``create`` methods that can be wired onto a
+    mock ``AsyncOgxClient`` so that ``get_all_conversation_items`` and
+    ``_write_summary_marker`` (and ``append_turn_items_to_conversation``) work
+    without patching.
```
max-svistunov, line 597: The docstring says the store exists so that `_write_summary_marker` works "without patching", but `_patch_write_summary_marker` (test file, line 89) patches it in every test, and the fake variant rebuilds the marker item by hand:

    marker_item = {"type": "message", "role": "user",
                   "content": [{"type": "input_text", "text": f"{MARKER_SENTINEL} {text}"}]}

That duplicates the construction at `src/utils/conversation_compaction.py:273-289`, so a change to the real marker format would leave these tests passing. The fake store's `create(conversation_id, *, items, **kwargs)` signature matches the real call site, so the patch can be dropped entirely - asserting on the store contents is both simpler and a stronger check than `mock_write_marker.assert_awaited_once()`.

tests/integration/endpoints/test_conversation_compaction_integration.py
```diff
@@ -0,0 +1,2465 @@
...
+        mock_summarize.assert_awaited_once()
+        mock_write_marker.assert_awaited_once()
+
+        agent_params = mock_query_agent.build_agent_mock.call_args[0][1]
+        assert agent_params.omit_conversation is True
+        assert isinstance(agent_params.input, list)
+
+        input_texts = [getattr(m, "content", "") for m in agent_params.input]
+        assert any("condensed earlier turns" in t for t in input_texts)
+        assert input_texts[-1] == "What else can you help with?"
```
max-svistunov: Nothing in the suite asserts that the compacted turn is written back to the conversation. In compacted mode the `conversation` parameter is dropped, so the service has to append the turn itself (`append_turn_items_to_conversation`, `src/utils/agents/query.py:279`). If that regresses, conversation history silently stops growing - no exception, no failing test, and the next request's recent-turn buffer is empty.

`mock_conversation_store` makes this a two-line check: after the handler returns, assert the store gained the user query and the assistant output.

tests/integration/endpoints/test_conversation_compation_integration.py
```diff
@@ -0,0 +1,2465 @@
...
+            query_endpoint_handler(
+                request=test_request,
+                query_request=QueryRequest(
+                    query="What is RHEL?", conversation_id=EXISTING_CONV_ID
+                ),
+                auth=test_auth,
+                mcp_headers={},
+            )
+        )
+        await asyncio.sleep(0.05)
```
max-svistunov: This assertion can pass without proving anything. If task 2 has not been scheduled far enough within the 50 ms window, `task2.done()` is False and `task2_entered` is unset for reasons unrelated to the lock - the test cannot tell "blocked on the lock" from "not started yet".

Suggest a deterministic signal instead of the sleep: have the task-2 path set an event immediately before it enters `apply_compaction`, and await that. Same pattern at lines 1153, 1755 and 2348.

tests/integration/conftest.py
```diff
@@ -827,6 +920,29 @@ def mock_ogx_client_fixture(
     yield mock_client
 
 
+@pytest.fixture(name="mock_conversation_store")
```
max-svistunov, line 1055: `tests/integration/README.md` documents every other shared fixture under "Common Fixtures" / "Mocking Fixtures". Worth adding `mock_conversation_store` and `InMemoryConversationStore` there - they are the entry point for any future stateful-conversation test, and the README is where contributors look first.

tests/integration/endpoints/test_conversation_compation_integration.py
Jazzcort: You're right! Thanks for the catch! Will add a deterministic check to make sure that task2 has started before checking `task2.done()` and `task2_entered.is_set()`

tests/integration/conftest.py
Jazzcort: will do!

Jazzcort:
@max-svistunov Need a quick second eye and we can merge this! 😁 
Those e2e test failures are definitely not caused by this PR since it only touch the test files.

tests/integration/endpoints/test_conversation_compation_integration.py
```diff
@@ -0,0 +1,3114 @@
...
+    cache_patch_target: str,
+    items: list[Any],
+) -> tuple[Any, AsyncMockType, AsyncMockType]:
+    """Set up mocks for recursive fold tests.
+
+    Creates a mock cache pre-loaded with two existing summaries whose combined
+    token count, when a third summary is added by compaction, exceeds the
+    compaction threshold — triggering ``_maybe_persist_fold``.
+
+    Each existing summary has ``token_count=8`` (total 16). The new summary
```
max-svistunov: The docstring says the existing summaries have `token_count=8` (total 16) and the new summary adds `token_count=6` (total 22), but the code below uses `1`, `1` and `1000`. The test still triggers the fold, because 1000 exceeds `200 * 0.1`, but the explanation no longer matches the values. Suggest updating the docstring to the numbers actually used.

max-svistunov (approved):
@Jazzcort LGTM, thanks! 👍 Nit: the SSE ordering test still accepts an `error` event (coderabbit's comment) -- worth dropping it before merge.

The other comment is written by Claude, it's less major, I checked it, it's valid.

Jazzcort:
@max-svistunov Thanks for the catch! I was moving the numbers around and forgot to update the docstring. You're right that we should take "error" off from the happy-path SSE set. I've addressed them all. 😁

Jazzcort:
@tisnik This PR is good to be merged! Since this PR only touched the test module, the failing E2E tests were not causing by this PR.

tests/integration/endpoints/test_conversation_compaction_integration.py
```diff
@@ -0,0 +1,3114 @@
+"""Integration tests for conversation compaction in query, A2A, streaming, and responses."""
```
radofuchs: please split this file into several smaller ones. Like this it is very hard to review without losing track of what is and is not actually being tested here

radofuchs (changes requested):

Jazzcort:
@radofuchs Didn't know it would grow into such a large module when I started. 😞 I've broken it down. Let me know if there are any changes needed.

tests/integration/endpoints/test_compaction_query.py
```diff
@@ -0,0 +1,641 @@
...
+        assert isinstance(agent_params.input, list)
+
+        input_texts = [getattr(m, "content", "") for m in agent_params.input]
+        assert len(input_texts) == 2
+        assert not any("question one" in t for t in input_texts)
+        assert not any("answer one" in t for t in input_texts)
+        assert not any("question two" in t for t in input_texts)
+        assert not any("answer two" in t for t in input_texts)
+        assert any(DEFAULT_SUMMARY_TEXT in t for t in input_texts)
+        assert input_texts[-1] == "What else can you help with?"
```
radofuchs, line 139: this assert is twice in this same test

tests/integration/endpoints/test_compaction_query.py
```diff
@@ -0,0 +1,641 @@
...
+
+        items_from_store = await collect_items(mock_conversation_store, CONV_ID_LLAMA)
+        assert len(items_from_store) == 7
+        expected = items + [
+            marker(DEFAULT_SUMMARY_TEXT),
+            msg("user", "What else can you help with?"),
+            msg("assistant", DEFAULT_MODEL_RESPONSE),
+        ]
+        assert verify_store_content(items_from_store, expected)
```
radofuchs, line 226: remove this part, it is only retesting the same thing as previous scenario

tests/integration/endpoints/test_compaction_query.py
```diff
@@ -0,0 +1,641 @@
...
+        mock_summarize.assert_awaited_once()
+        assert_marker_count(mock_conversation_store, CONV_ID_LLAMA, 1)
+
+        agent_params = mock_query_agent.build_agent_mock.call_args[0][1]
+        assert agent_params.omit_conversation is True
+        assert isinstance(agent_params.input, list)
+
+        input_texts = [getattr(m, "content", "") for m in agent_params.input]
+        assert len(input_texts) == 2
+        assert not any("question one" in t for t in input_texts)
```
radofuchs, line 134: these negative asserts are not necesary here as they do not really test anything. When you test the lengts and the content of the list, then testing that some other elements are not there makes no sense

Applies to all relevant scenarios below as well

tests/integration/endpoints/test_compaction_query.py
```diff
@@ -0,0 +1,641 @@
...
+        Verifies:
+        - summarize_chunk is NOT called (under threshold)
+        - Agent receives compacted params with summary from the marker,
+          recent messages, and the new query
+        """
+        _ = mock_ogx_client
+
+        enable_compaction(
+            test_config,
+            context_window=1_000_000,
```
radofuchs, line 251: set this window just above the amount of tokens spent so that it is conclusively verified that the compaction does not trigger.

tests/integration/endpoints/test_compaction_query.py
```diff
@@ -0,0 +1,641 @@
...
+        assert len(items_from_store) == 7
+        expected = items + [
+            marker(DEFAULT_SUMMARY_TEXT),
+            msg("user", "What else can you help with?"),
+            msg("assistant", DEFAULT_MODEL_RESPONSE),
+        ]
+        assert verify_store_content(items_from_store, expected)
+
+    @pytest.mark.asyncio
+    async def test_query_compaction_existing_marker_no_new_summarization(
```
radofuchs, line 229: what if there are more markers? I would want to see a test like this instead of this one

tests/integration/endpoints/test_compaction_query.py
```diff
@@ -0,0 +1,641 @@
...
+
+        Verifies:
+        - No summarization or marker write
+        - Agent receives normal (non-compacted) params
+        """
+        _ = mock_ogx_client
+
+        enable_compaction(
+            test_config,
+            context_window=1_000_000,
```
radofuchs, line 324: same as before, this context window is too huge for no reason

tests/integration/endpoints/test_compaction_query.py
```diff
@@ -0,0 +1,641 @@
...
+            threshold_ratio=0.5,
+            buffer_turns=4,
```
radofuchs: why set the buffer to 4 here? Looks like it is trying to hide something here

radofuchs (changes requested):
please apply all my comments to all files

tests/integration/endpoints/test_compaction_responses.py
```diff
@@ -0,0 +1,777 @@
...
+        items_from_store = await collect_items(mock_conversation_store, CONV_ID_LLAMA)
+        assert len(items_from_store) == 7
+        expected = items + [
+            msg("user", "Any updates?"),
+            msg("assistant", DEFAULT_MODEL_RESPONSE),
+        ]
+        assert verify_store_content(items_from_store, expected)
+
+    @pytest.mark.asyncio
+    async def test_responses_compaction_small_conversation_no_compaction(
```
snuryyeva, line 409: I think we should also add one more negative scenario here, where we hit responses APi without conversationID, which should behave in a similar was as this scenario and not trigger summarize_chunk

tests/integration/endpoints/test_compaction_streaming.py
```diff
@@ -0,0 +1,882 @@
...
+        assert not task2.done()
+        assert not task2_entered.is_set()
+
+        release.set()
+        await asyncio.gather(task1, task2)
+
+        assert task2_entered.is_set()
+
+    @pytest.mark.asyncio
+    async def test_streaming_compaction_sse_event_ordering(
```
snuryyeva: I think this test and est_streaming_compaction_emits_start_and_compaction_sse_events can be combined into one

snuryyeva (changes requested):

tests/integration/endpoints/test_compaction_query.py
Jazzcort: So, from my point of view, each test exercises a complete request flow with a distinct configuration, and verifying all observable side effects (agent params and store state) is a right thing to do. 

If we verify the store in one test but skip it in another that runs the same endpoint, a future reader has to wonder whether the omission is intentional or an oversight. Consistency removes that ambiguity.

The store outcome happens to be the same today, but each test covers a distinct config path end-to-end. Keeping all tests structurally consistent — always verifying both agent params and store state — makes the suite easier to read and more resilient to future changes. The cost is a few lines.

tests/integration/endpoints/test_compaction_query.py
Jazzcort: The multi-marker path is actually already exercised by test_query_conversation_compaction_additive_summarization. This test isolates a different concern — existing marker without re-triggering summarization. I think adding more markers here doesn't not really increase the coverage.

What do you think?

tests/integration/endpoints/test_compaction_query.py
Jazzcort: Setting a large number is just to be defensive that the further compaction will not be triggered. And, it also kinda show the intend "I don't want the compaction to happen" to the future readers, so they don't have to print some output there to verify that the compaction will not happen.

I think the point here is that (mock_summarize.assert_not_called()) is what verifies the behavior. The proximity of the threshold here does not really matter that much.

tests/integration/endpoints/test_compaction_query.py
Jazzcort: ~~I know your points and I was thinking about that too. I'll remove them~~
I changed my mind on what I said above.

I see your point that the negative assertions are logically redundant given the length and content checks. That said, I'd prefer to keep them here — they serve more as documentation of the compaction contract than as functional assertions. A future reader seeing `assert not any("question one" in t ...)` immediately understands "old turns must not leak into the compacted input" without having to mentally derive that from the length + positive checks. Since these assert the absence of specifically the items that were present before compaction, I think they pull their weight as intent markers. Happy to hear your thoughts though.

tests/integration/endpoints/test_compaction_query.py
Jazzcort: What do you mean for "twice" here? 🤔

tests/integration/endpoints/test_compaction_query.py
Jazzcort: You're right. Actually buffer_turn does not really matter here, I'll just remove it so it uses the default value.

tests/integration/endpoints/test_compaction_responses.py
Jazzcort: Nice catch! I'll add another test for this. Note that this behavior only applies to the `/responses` endpoint.

tests/integration/endpoints/test_compaction_streaming.py
Jazzcort: This makes sense to me. I'll move the subset checking into est_streaming_compaction_emits_start_and_compaction_sse_events and just remove this one. Thanks for the catch.

Jazzcort:
@radofuchs I've addressed some comments and left some of my thoughts on those ones that I'm not quite agree with. Overall, those disagreement are more philosophical other than functional, which means it does not affect the test coverage.

@snuryyeva Thanks for the suggestions. I've addressed them as well.

tests/integration/endpoints/test_compaction_query.py
radofuchs: I agree with your statement that this is defensive, but on the other hand it might seem as hiding a fact that if you set the number to anything below this it might actually trigger

tests/integration/endpoints/test_compaction_query.py
radofuchs: no, it does not increase the coverage. My idea is more in the normal working environment where there are more markers, the logic of the test remains almost unchanged

snuryyeva (approved):
LGTM

radofuchs (approved):
LGTM

PR #2434, opened 2026-08-13
LCORE-1675: Documentation for conversation compaction

Jazzcort:
## Description

Document the conversation compaction feature across OpenAPI spec, configuration guide, architecture overview, and query endpoint docs. Add context_status field ("full"/"summarized") to QueryResponse documentation and to the end SSE event for streaming query. Create comprehensive user guide at docs/user_doc/conversation_compaction.md with configuration examples, behavior details, and FAQ.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [x] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Assisted-by: claude-opus-4.6

- Related Issue 
- Closes [LCORE-1675](https://redhat.atlassian.net/browse/LCORE-1675)

## Testing
Read and double check maybe

## Documentation

- Added a user guide for conversation compaction, covering configuration, behavior, prerequisites, and FAQs.
- Added navigation links to the conversation compaction documentation.
- Documented conversation context status in query responses, including streaming events and examples.
- Updated architecture and design documentation with compaction flow, endpoint integration, configuration, and concurrency details.
- Clarified configuration options and context-budget behavior.
- Clarified token-estimation limitations for large tool-result payloads.


Jazzcort:
@max-svistunov Probably need a second eye on this. 😁

max-svistunov (commented):
@Jazzcort Thanks, this is good 👍 Two things:

* (this is primarily my fault) The `openapi.md` row says `context_status` is on `StreamingQueryResponse`, but that class is documentation-only with an empty body. The streaming surface is `EndEventData`, which is where LCORE-1573 put it. `query_endpoint.md` and the guide are both correct, so it's only that row (and the PR description if it matters).

* 1573 merged yesterday, so `openapi.json` on main now has `context_status` but `openapi.md` and `query_endpoint.md` don't. Thsi PR fixes that, so can you rebase please?

Again, other comments are written by Claude, they are less major, and I checked that they are OK.

docs/user_doc/conversation_compaction.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Frequently asked questions
+
+**Does compaction lose information?**
+
+Compaction summarizes older turns, so fine-grained details from early in the conversation may be condensed. The full original conversation history remains stored in OGX and is retrievable via the conversations API. The LLM simply receives a summary instead of the full transcript for inference.
+
+**Does compaction use extra tokens?**
+
+Yes. The summarization step requires an additional LLM call, which consumes tokens. These tokens are counted against the user's quota. The trade-off is that the conversation can continue instead of failing with HTTP 413.
```
max-svistunov, line 159: "These tokens are counted against the user's quota" does not match the implementation. `summarize_chunk` calls `client.responses.create` directly (`src/utils/compaction.py:200`) and discards the returned usage - `ConversationSummary.token_count` is a fresh tiktoken count of the summary text, not the LLM's reported usage. Quota consumption runs only off the turn's `token_usage` (`src/utils/query.py:300`).

So the summarization call is invisible both to quota and to the `input_tokens` / `output_tokens` reported in the response. That is the opposite of what an operator would assume from this FAQ entry, so it is worth stating explicitly rather than leaving as-is.

docs/user_doc/conversation_compaction.md
```diff
@@ -0,0 +1,171 @@
...
+
+After compaction, the service takes ownership of the context window. The full original conversation history remains stored in OGX for auditing and retrieval, but only the compacted view is sent to the LLM for inference.
+
+### Affected endpoints
+
+| Endpoint | Compaction behavior |
+|---|---|
+| `POST /v1/query` | Blocking compaction before inference. Response includes `context_status`. |
+| `POST /v1/streaming_query` | Compaction runs inside the SSE stream. A `compaction` event is emitted before tokens begin. The `end` event includes `context_status`. |
+| `POST /v1/responses` | Compaction runs silently (no `context_status` in response). The `/v1/responses` endpoint follows the OpenAI Responses API specification and does not add custom fields. |
```
max-svistunov, line 24: Compaction on `/v1/responses` is conditional, not unconditional. `src/app/endpoints/responses.py:709` gates it on `store` **and** `conversation` **and** `not previous_response_id`, so a stateless request, or one chained via `previous_response_id`, never compacts at all.

Worth saying so here and in the ARCHITECTURE table (line 432) - that is the case an integrator is most likely to hit, and "compaction runs silently" currently implies it always runs.

docs/user_doc/config.md
```diff
@@ -224,6 +224,51 @@ Attributes:
 | buffer_turns     | integer | Number of recent turns to keep verbatim.                                                                    |
 | buffer_max_ratio | number  | Maximum fraction of context window the buffer zone can occupy, regardless of buffer_turns.                  |
 
+### How to enable conversation compaction
```
max-svistunov, line 217: Two problems with placing this block here.

`docs/user_doc/config.{json,md,html}` are maintained by hand as a set (no Makefile target, script or workflow generates them). `config.html` already carries the compaction schema, and this change leaves it behind, so the published HTML and the Markdown diverge.

Second, the new prose sits inside the generated-schema region, between the `CompactionConfiguration` table and `## Configuration`. Whenever config.md is next regenerated from the schema, it gets dropped.

Since `conversation_compaction.md` already carries the same YAML examples almost verbatim, the cleanest fix is to drop this block and keep only the cross-link at the end of the section.

docs/user_doc/conversation_compaction.md
```diff
@@ -0,0 +1,171 @@
...
+
+## When compaction is disabled
+
+When compaction is disabled (the default), requests that cause the conversation history to exceed the model's context window will fail with HTTP 413 (Prompt Too Long). Clients must manage conversation length themselves, for example by starting new conversations or deleting old ones.
+
+## Frequently asked questions
+
+**Does compaction lose information?**
+
+Compaction summarizes older turns, so fine-grained details from early in the conversation may be condensed. The full original conversation history remains stored in OGX and is retrievable via the conversations API. The LLM simply receives a summary instead of the full transcript for inference.
```
max-svistunov, line 155: "The full original conversation history remains stored in OGX and is retrievable via the conversations API" is true, but incomplete in a way that matters to anyone rendering that history.

Compaction writes each summary as a synthetic `user` message prefixed with `[lightspeed:compaction-summary]` (`src/utils/conversation_compaction.py:273`), and nothing in `conversations_v1.py`, `conversations_v2.py` or `utils/conversations.py` filters marker items out. So `GET /v1/conversations/{id}` returns those markers as though the user had typed them. Worth a sentence here so UI consumers know to expect them.

docs/user_doc/conversation_compaction.md
Jazzcort: Thanks for the catch! The token usage during compaction is only used during fold process. I'll update this!

docs/user_doc/conversation_compaction.md
Jazzcort: Will update it!

docs/user_doc/config.md
Jazzcort: I'll remove this block, and only leave the cross-link.

docs/user_doc/conversation_compaction.md
Jazzcort: Agree! We should add these details so it would be less confusing when people is checking the conversation.

Jazzcort:
@max-svistunov Should be good! Might need a quick second eye from you. 😁

max-svistunov (commented):
@Jazzcort Thanks, almost there 👍 Only one thing left from my side -- the openapi.md row still says `context_status` is on `StreamingQueryResponse`, but in the response it's in the `end` event (`EndEventData`). The open coderabbit threads look valid to me too, esp. the degrading guard one (the budget is `buffer_max_ratio` of the window, not window minus summary minus query). The build-pr failure is a dnf/infra error, a re-run should fix it.

Jazzcort:
@max-svistunov Thanks for the catch! I've addresses all of them! 😁

Jazzcort:
@max-svistunov The last coderabit comment is valid. We do not include the tokens of non-message item, so I modified some statements to address that.

max-svistunov (commented):
@Jazzcort Thanks, the openapi row looks good now 👍 One thing still left though -- the degrading guard section says the budget is the context window minus the summary minus the new query, but in the code it's `buffer_max_ratio` of the window (`budget = int(context_window * buffer_max_ratio)`). The other open coderabbit threads (threshold wording at the top of the guide, the note about disabling compaction mid-conversation, the diagram order) are still there too.

Jazzcort:
@max-svistunov Thanks for the catch! I've addressed them.

max-svistunov (approved):

tisnik (approved):
LGTM

PR #2551, opened 2026-08-26
LCORE-3743: Fix vector-store file upload OOM and add concurrency limits

JslYoon:
## Jira
https://redhat.atlassian.net/browse/LCORE-3743

## Summary
- Library mode's `client.files.create()` required a `BytesIO` instance (`ogx.core.library_client._handle_file_uploads`) and re-buffered it into a second in-memory copy on top of the caller's own buffer, doubling peak memory for large uploads. Fixed by bypassing `client.files.create()` in library mode via a new `AsyncOgxClientHolder.upload_file()` that calls the Files provider directly — cuts peak memory 2x → 1x per upload (verified empirically against the pinned `ogx==1.0.2`). Service/HTTP mode is unchanged (already streams via httpx).
- Added configurable concurrency caps (`service.max_concurrent_file_uploads` / `service.max_concurrent_vector_store_attaches`, both default 5) on `POST /v1/files` and `POST /v1/vector-stores/{id}/files`, since concurrent requests otherwise multiply memory linearly with no bound. Excess requests get `429 Too Many Requests`.
- Delete the source file once it's successfully attached to a vector store — nothing downstream (citations, retrieval) ever re-reads it, and `localfs` has no active TTL reaper, so uploads would otherwise accumulate on disk forever. Skipped on a failed attach so retries still work.
- Fixed a filename-normalization inconsistency found while refactoring: the `.txt` extension appended for extensionless uploads wasn't applied in library mode (the Files provider reads `file.filename` directly, bypassing the endpoint's locally-computed name).
- Documented the new config knobs in `docs/user_doc/rag_guide.md`.

## Test plan
- [x] `make format` / `make verify` (black, pylint 10.00/10, pyright, ruff, pydocstyle) — all clean
- [x] `make test-unit` — 3365 passed, 1 skipped, 90.62% coverage
- [x] New unit tests: `AsyncOgxClientHolder.upload_file()` (library mode, service mode, filename fix, uninitialized-impls error), semaphore 429 + logging for both endpoints, delete-after-attach (completed/failed/delete-fails-but-response-still-succeeds)
- [ ] e2e: not added — no existing multipart file-upload step definition in `tests/e2e/features/steps/`, and verifying new e2e coverage requires a live service with real inference/embedding credentials not available in this environment. Flagging for follow-up rather than shipping unverified step definitions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added configurable concurrency limits for file uploads and vector-store attachments, defaulting to 5 each.
  * Added clear `429 Too Many Requests` responses when concurrency limits are reached.
  * Added automatic cleanup of source files after successful vector-store attachment.
  * Added file-upload support across supported client modes.

* **Documentation**
  * Documented concurrency limits, `429` behavior, memory considerations, and file cleanup.


JslYoon:
Thanks for flagging this — verified both points against the actual pinned `ogx==1.0.2` code before fixing.

**Delete-after-attach reuse issue: confirmed, fixed in 3b57354.** `openai_attach_file_to_vector_store` re-fetches the file via `openai_retrieve_file()` on every call (`openai_vector_store_mixin.py:1265`), so a second attach of the same `file_id` after our delete does fail — though the failure mode is `200 OK` with `status: "failed"` in the body, not an HTTP 404 (lightspeed-stack doesn't expose a standalone `GET /v1/files/{file_id}` route to 404 on). Fixed by making it opt-in: added `service.delete_file_after_vector_store_attach` (default `false`), so files stay reusable across multiple vector stores by default, matching the OpenAI Files API. Deployments that only ever attach a file once can opt in.

**Concurrency settings requiring a restart:** confirmed true, but not something this PR introduces. `configuration.configuration` is loaded exactly twice in the whole codebase (CLI + startup) and is never hot-reloaded anywhere — every existing `ServiceConfiguration` field (`workers`, `auth_enabled`, `host`, `port`, ...) already requires a restart to change. The `@lru_cache`-based semaphores are consistent with that existing behavior, not a new limitation.

New tests added covering the reuse fix (`test_add_file_to_vector_store_keeps_file_reusable_by_default`) and the opt-in enabled path. Full suite green (3366 passed), lint/type-check clean.

tisnik (approved):
LGTM

PR #2592, opened 2026-09-02
LCORE-3883: persist compacted-mode turns on query, streaming and A2A

max-svistunov:
## Description

Fixes LCORE-3883. In compacted mode the `conversation` parameter is dropped from the OGX call, so OGX does not store the turn and lightspeed-stack must append it itself. That wiring was lost: `29bd62d7` removed the A2A call site while refactoring the executor onto pydantic-ai agents, and `16c17d44` removed the query and streaming call sites along with the deprecated helpers they lived in. `/v1/responses` was unaffected because it has its own `_append_previous_response_turn`.

The effect is silent data loss: once a conversation enters compacted mode, no further turn is recorded. History stops growing, and because the recent-turn buffer is derived from the conversation items after the last marker, the model loses all context after the compaction point. Nothing raises. It was masked until LCORE-3582 landed, since compacted requests previously failed with HTTP 500 before reaching the point where the turn would be stored.

Restoring the call sites alone is not sufficient. `TurnSummary.output_items` — the field the persistence writes — is no longer populated on any success path either; the only writer left is the blocked-moderation branch. Wiring without it would append the user message with an empty assistant reply.

Rather than reconstruct output items by mapping pydantic-ai messages back to the OGX shapes, they are captured where the genuine objects already exist: `OgxResponsesModel` now records `response.output` from the non-streaming call and from the `response.completed` event on the streaming path. A compacted turn is therefore stored exactly as OGX itself would have stored it.

Keeping the `conversation` parameter so that OGX persists the turn was tested against a local OGX 1.0.2 stack and rejected. OGX does honour the explicit input and does not inject conversation history, but it persists the whole input list, so every replayed buffer turn is duplicated into the conversation on each request (a one-turn buffer grew a conversation from 2 to 7 items in a single request, and the duplication compounds because the buffer is read back from those same items).

Notes for reviewers:

1. The streaming success path shares the existing interrupt `persist_guard`, so a turn is persisted at most once if an interrupt fires while the stream is finishing.
2. A persistence failure is logged rather than failing an otherwise successful stream — the client already has its answer, and the cost is that the next request loses that turn. Happy to invert this if you would rather it surface.
3. The comment at `streaming_query.py:491` claiming the shared generator handles compacted-turn storage needed no edit: this change makes it true again.
4. Tool-call and tool-result items are now preserved in compacted history, because the captured items are the real OGX output rather than a flattened assistant message.

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Related Issue # LCORE-3582
- Closes # LCORE-3883

## Testing

Reproduce the OGX persistence behaviour that motivates the design (requires a local OGX stack):

1. Start OGX: `OPENSSL_CONF=/dev/null OPENAI_API_KEY=... ogx stack run tests/e2e/configs/run-ci.yaml --port 8321`
2. Create a conversation, seed two turns, then `POST /v1/responses` with `conversation` set and an explicit input list containing a summary plus one replayed turn pair plus a new query.
3. Expected, and observed: the model answers from the explicit input (conversation history is not injected), but the conversation grows by 5 items rather than 2 — the replayed pair is duplicated. This is why the fix appends the turn explicitly instead of letting OGX store it.

Run the unit tests covering the three endpoints:

```
uv run python -m pytest tests/unit/utils/agents/ tests/unit/pydantic_ai_lightspeed/ tests/unit/app/endpoints/test_a2a.py -q
```

Result: `346 passed`. Full unit run across the touched areas: `1268 passed`.

The new tests assert against the conversation items that actually land rather than the arguments of a mocked helper. Confirmed non-vacuous: reverting `src/utils/agents/streaming.py` fails `test_compacted_success_appends_turn_to_conversation`, and reverting `src/utils/agents/query.py` fails `test_compacted_success_appends_turn_with_captured_output`.

Linters:

```
uv run make format && uv run make verify
```

black, ruff, pylint, pyright and docstyle pass. `check-types-src` reports `Success: no issues found in 223 source files`. `check-types-tests` reports 38 errors, all pre-existing on `main` and none in the files touched here.

## Summary by CodeRabbit

* **New Features**
  * Compacted A2A, query, and streaming turns are now saved to the conversation after completion.
  * Structured model output is captured and included when persisting completed turns.
  * Persistence is protected against duplicate writes and does not interrupt successful responses if saving fails.

* **Bug Fixes**
  * Prevented compacted turns from being omitted from conversation history.
  * Non-compacted requests continue without adding unnecessary conversation entries.

### DIFF
--- a/src/app/endpoints/a2a.py
+++ b/src/app/endpoints/a2a.py
@@ -60,9 +60,14 @@ from configuration import configuration
 from constants import MEDIA_TYPE_EVENT_STREAM
 from log import get_logger
 from models.api.requests import QueryRequest
+from models.common.responses.responses_api_params import ResponsesApiParams
 from models.config import Action
 from utils.agents.error_handler import map_agent_inference_error
-from utils.conversation_compaction import apply_compaction_blocking
+from utils.conversation_compaction import (
+    CompactionResult,
+    apply_compaction_blocking,
+    store_compacted_turn,
+)
 from utils.mcp_headers import McpHeaders, mcp_headers_dependency
 from utils.otel_tracing import (
     SpanAttributes,
@@ -71,7 +76,7 @@ from utils.otel_tracing import (
     anonymize_value,
     set_span_attributes,
 )
-from utils.pydantic_ai_helpers import build_agent
+from utils.pydantic_ai_helpers import build_agent, captured_output_items
 from utils.query import extract_provider_and_model_from_model_id
 from utils.responses import prepare_responses_params
 from utils.suid import normalize_conversation_id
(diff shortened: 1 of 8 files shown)


tisnik (approved):
LGTM

PR #2596, opened 2026-09-02
LCORE-: Fix missing refusal message in conversation/v2 when using streaming query

Jazzcort:
## Description

When streaming completes without partial tokens, the LLM response was not being captured in the turn summary, causing refusal messages to be lost in conversation/v2 history.

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Related Issue #
- Closes #

## Testing
It's better to verify the bug first.
1. Configure a shield - question validity should be the easiest to set up
2. Ask a question that will be blocked by the shield through `/streaming_query`
3. Check the conversation history with the conversation_id from the response in `/conversation/v2`
4. The assistant message should be an empty string

Now apply the patch and redo the above steps, but this time the refusal message should show up in `conversation/v2`

## Summary by CodeRabbit

* **Bug Fixes**
  * Turn summaries now capture the final response when no streamed tokens were received, including content-filter refusals and other non-streamed responses.

### DIFF
--- a/src/utils/agents/streaming.py
+++ b/src/utils/agents/streaming.py
@@ -516,6 +516,11 @@ def _(
     )
     state.chunk_id += 1
     state.turn_summary.next_chunk_id = state.chunk_id
+
+    # Fix the missing refusal message issue in conversation/v2
+    if not state.turn_summary.partial_tokens:
+        state.turn_summary.llm_response = final_text
+
     return payload
 
 


src/utils/agents/streaming.py
asimurka, line 506: I would rather do the assignment here. If the shield capability is correctly implemented it should raise "content_filter" finish reason. In that case the `llm_response` is indeed not assigned.

asimurka (changes requested):
I don't think this is the right way how to fix this issue despite it is valid. If partial tokens list is empty, it means that the "short circuit" was executed - that's correct. But it is just a side effect. Fix it at the correct place.

Jazzcort:
@asimurka So I first tried to make question validity to respond with `finish_reason="content_filter"` and it kinda cause a bunch of more problems - both `query` and  `streaming_query` will raise 500 which I think it not right. At least for "content_filter", it just means the model refuse to respond due to its policy but the whole system is actually working fine. Raise a 500 there is semantically wrong from my opinion.

Another thing is that after some thinking, I feel like our own shield capacity should not step into the model's behavior. If we reply with `finish_reason="content_filter"` for all the shield rejection, it kind of blur the line between how the model behavior and our application-level guard. 

Therefore, my proposal is the following:
1. When the input is reject by a shield, we still respond `finish_reason="stop"`
2. We take advantage of the `metadata` field of `GraphAgentState` which will then be exposed to `AgentRunResult.metadate` so we can send our own data in between agent steps and surface it to the result.

The draft might look like this. What do you think?
```python
@dispatch_stream_event.register
def _(
    event: AgentRunResultEvent,
    state: AgentTurnAccumulator,
) -> Optional[StreamEventPayload]:
    """Handle final run result event and emit completion payload.

    Args:
        event: Final run result event.
        state: Mutable accumulator for stream processing.

    Returns:
        Completion stream payload.
    """
    state.run_result = event.result
    state.turn_summary.id = state.run_result.response.provider_response_id or ""
    if state.run_result.response.finish_reason == "content_filter":
        provider_details = state.run_result.response.provider_details or {}
        final_text = provider_details.get("refusal") or DEFAULT_REFUSAL_RESPONSE

        state.turn_summary.llm_response = final_text

    else:
        final_text = state.run_result.response.text or "".join(state.text_parts)

        lightspeed_stack_metadata = (state.run_result.metadata or {}).get(
            "_lightspeed_stack", {}
        )

        if "refusal" in lightspeed_stack_metadata:
            final_text = lightspeed_stack_metadata["refusal"]
            state.turn_summary.llm_response = final_text

    payload = TurnCompleteStreamPayload.create(
        chunk_id=state.chunk_id,
        token=final_text,
    )
    state.chunk_id += 1
    state.turn_summary.next_chunk_id = state.chunk_id

    return payload

```

Btw, both `OpenAIResponsesModel` and `OpenAIChatCompletionModel` put refusal message at "refusal" field of `provider_details` I think.

Jazzcort:
@asimurka I was reading your [Conversation API](https://docs.google.com/document/d/1XnlS1APc7mfhkQ2X9hrTlJ6HtQMrvIGsDqqhZY67S9s/edit?tab=t.0). Since we want the shield blocking message to persist, we can not set `finish_reason` to content_filter because it will cause a 5XX error. Or, we make `finish_reason="content_filter"` not raise an error. 😁 What do you think?

PR #2609, opened 2026-09-03
LCORE-1822: Enable OpenTelemetry delivery E2E test with mock OTLP collector

anik120:
## Description

Implements the Behave step definitions for the previously-skipped OpenTelemetry E2E scenario and turns the test on, so CI now verifies end-to-end that the Lightspeed Core Stack actually delivers telemetry to an OTLP backend.

The scenario asserts that a `responses` request carrying a `safety_identifier` marker results in that marker reaching a collector. Two pieces were needed to make it real:

1. **A mock OTLP/HTTP collector** to receive and assert on exports.
2. **Instrumentation** so the marker (`safety_identifier`) is actually emitted on a span — previously it was only forwarded to the model provider, so the scenario could never pass and was tagged `@skip`.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Related Issue #
- Closes #

## Testing

## Summary by CodeRabbit

- **New Features**
  - OpenTelemetry export is now enabled by default for supported deployments.
  - Telemetry can be sent to a configurable OTLP/HTTP endpoint.
  - Response traces now include a caller-provided safety identifier when available.

- **Testing**
  - Added an integrated mock OpenTelemetry collector for end-to-end validation.
  - Expanded coverage to verify telemetry delivery and safety-identifier recording.


tests/e2e/features/steps/opentelemetry.py
```diff
@@ -0,0 +1,223 @@
...
+        capture_output=True,
+        text=True,
+        timeout=timeout,
+        check=False,
+    )
+    if result.stdout:
+        print(result.stdout, end="")
+    if result.returncode != 0:
+        print(result.stderr, end="")
+        raise AssertionError(f"`{' '.join(cmd)}` failed with code {result.returncode}")
```
snuryyeva: Do we need to have compose logic here if we are already have set up added in docker-compose.yaml and docker-compose-library.yaml

tests/e2e/features/steps/opentelemetry.py
```diff
@@ -0,0 +1,223 @@
...
+    collector, which are not deployed on Prow/OpenShift. Returns True when the
+    scenario was skipped so the caller can return early.
+    """
+    if is_prow_environment():
+        context.scenario.skip(
+            "OpenTelemetry delivery scenario requires Docker Compose and the "
+            "mock OTEL collector, which are not deployed on Prow/OpenShift."
+        )
+        return True
+    return False
```
snuryyeva: we can handle this with tag. we have @skip-in-prow tag that can be added to the feature file itself, so the feature does not run in prow.

tests/e2e/features/opentelemetry.feature
snuryyeva: these two steps are both doing the same verification, they poll to check if marker is received by the mock. we could remove one step.

tests/e2e/mock_otel_collector/server.py
```diff
@@ -0,0 +1,166 @@
...
+- ``POST /v1/logs``    : Receive an OTLP log export (log records / events).
+- ``POST /v1/metrics`` : Receive an OTLP metric export.
+  Each returns ``200`` with an empty ``application/x-protobuf`` body, which the
+  OTLP/HTTP exporter accepts as success.
+- ``GET /received``    : Return a JSON summary of everything buffered so far.
+  With ``?contains=<text>`` it reports whether that substring appears in any
+  buffered payload (raw-byte search; OTLP protobuf stores string fields as
+  UTF-8, so a plaintext marker embedded in an attribute value is found).
+- ``POST /reset``      : Clear the buffer (used at the start of a scenario).
+- ``GET /health``      : Liveness probe returning ``{"status": "ok"}``.
```
snuryyeva: for this e2e scenario we do not need to overcomplicate mock with so many different endpoints, I believe we should get away with /recieved, /reset and maybe /health ?

snuryyeva (changes requested):

tests/e2e/features/steps/opentelemetry.py
anik120: Okay so I removed the on-demand `mock-otel` bring-up. It's defined in both `docker-compose.yaml` and `docker-compose-library.yaml` with no profiles, so `docker compose up -d` already starts it; the step now just waits for it to be healthy and resets its buffer.

I kept one compose call: force-recreating lightspeed-stack in the "configure to export" step. That's required because the OTEL SDK only initializes when the entrypoint launches the app under opentelemetry-instrument (`OTEL_SDK_DISABLED=false`), and the exporter target is read from env baked in at container-creation — a docker restart won't apply new env, and we don't want OTEL enabled for every scenario. I moved it into a shared `force_recreate_compose_service` util and added an `after_feature` teardown that reverts the OTEL env and recreates the container so later scenarios aren't affected.

src/app/endpoints/responses.py
```diff
@@ -578,14 +578,18 @@ async def handle_responses_with_tracing(  # pylint: disable=too-many-locals
...
+    span_attributes: dict[str, Any] = {
+        SpanAttributes.USER_ID: anonymize_value(user_id),
+        SpanAttributes.INPUT: anonymize_value(input_text),
+        SpanAttributes.REQUEST_ATTACHMENTS_COUNT: attachments_count,
+    }
+    # safety_identifier is a caller-supplied, non-PII identifier, so it is
+    # recorded verbatim (not anonymized) when present.
+    if original_request.safety_identifier is not None:
+        span_attributes[SpanAttributes.SAFETY_IDENTIFIER] = (
+            original_request.safety_identifier
```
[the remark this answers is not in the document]
anik120: Keeping `safety_identifier` verbatim by design. It's OpenAI's documented, caller-supplied opaque identifier — the API docs explicitly recommend callers send a hash of the username/email rather than the raw value, i.e. it's intended to be non-PII. We record it unaltered so it stays correlatable across systems; USER_ID and INPUT remain anonymized via anonymize_value(). No change here.

anik120:
@snuryyeva thanks for the review! Addressed your comments PTAL

snuryyeva (approved):
LGTM

tests/e2e/features/steps/opentelemetry.py
```diff
@@ -0,0 +1,156 @@
...
+_MOCK_OTEL_PORT = os.getenv("E2E_OTEL_MOCK_PORT", "4318")
+MOCK_OTEL_CONTROL_BASE = f"http://{_MOCK_OTEL_HOST}:{_MOCK_OTEL_PORT}"
+
+# Delivery is asynchronous: the SDK batches spans before export. Poll generously.
+_DELIVERY_TIMEOUT_S = float(os.getenv("E2E_OTEL_DELIVERY_TIMEOUT_S", "45"))
+_DELIVERY_POLL_INTERVAL_S = 2.0
+_HEALTH_TIMEOUT_S = 30.0
+
+
+def _wait_for_mock_health() -> None:
```
radofuchs, line 54: this private method has no point, remove it, since it does the same thing as the healthcheck in the docker compose

tests/e2e/features/steps/opentelemetry.py
```diff
@@ -0,0 +1,156 @@
...
+@given("An OpenTelemetry service is running and listening for OTLP data")
+def otel_service_running(context: Context) -> None:
+    """Ensure the mock OTLP collector is healthy and its buffer is empty.
+
+    The ``mock-otel`` Compose service starts with the rest of the stack, so this
+    step only waits for it to report healthy (Docker health and its own control
+    API) and resets any previously buffered telemetry so the scenario starts
+    from a clean slate.
+    """
+    wait_for_container_health(MOCK_OTEL_SERVICE)
```
radofuchs, line 107: there are two healthchecks for the same thing, the only thing that should be realistically left in this step definition is the reset of mock collector

tests/e2e/features/steps/opentelemetry.py
```diff
@@ -0,0 +1,156 @@
...
+    # before_all sets OTEL_ANONYMIZATION_SECRET; keep any existing value.
+    os.environ.setdefault(
+        "OTEL_ANONYMIZATION_SECRET", "e2e-test-secret-do-not-use-in-production"
+    )
+
+    force_recreate_compose_service(
+        LIGHTSPEED_STACK_SERVICE,
+        is_library_mode=getattr(context, "is_library_mode", False),
+    )
+    wait_for_container_health(LIGHTSPEED_STACK_SERVICE)
```
radofuchs, line 137: we already have a behave step for lightspeed-stack restart, use that one instead

tests/e2e/utils/utils.py
```diff
@@ -515,6 +515,64 @@ def restart_lightspeed_stack_service(
...
+
+    Returns:
+        Absolute path to ``docker-compose-library.yaml`` in library mode,
+        otherwise ``docker-compose.yaml``.
+    """
+    name = "docker-compose-library.yaml" if is_library_mode else "docker-compose.yaml"
+    return absolute_repo_path(name)
+
+
+def force_recreate_compose_service(
```
radofuchs, line 532: please create a behave step deifiniton from this

tests/e2e/features/environment.py
```diff
@@ -39,10 +39,23 @@
...
     remove_config_backup,
     restart_container,
     switch_config,
+    wait_for_container_health,
+    wait_for_lightspeed_stack_http_ready,
+)
+
+# OTEL exporter env vars set by the OpenTelemetry feature's steps. They are
+# reverted after the feature so later scenarios do not inherit OTEL export.
+_OTEL_FEATURE_TAG = "OTel"
```
radofuchs, line 53: remove this, the tags are never meant to hide any kind of logic behind them

tests/e2e/features/environment.py
```diff
@@ -498,14 +511,42 @@ def _restore_config_after_feature_enabled() -> bool:
...
+    force-recreated (a plain ``docker restart`` would keep the OTEL-enabled
+    creation-time env), so later scenarios do not keep exporting to the mock
+    collector. No-op on Prow, where the feature is skipped and Compose is
+    unavailable.
+    """
+    if is_prow_environment():
+        return
+    for var in _OTEL_ENV_VARS:
+        os.environ.pop(var, None)
+    force_recreate_compose_service(
```
radofuchs, line 529: why do we actually need to disable this? I see no relevant reason in having to recreate the whole container just to get rid of the env variables

tests/e2e/features/steps/opentelemetry.py
```diff
@@ -0,0 +1,156 @@
...
+    """Enable OTEL export and recreate the service so the exporter is active.
+
+    The OTEL SDK is only initialized when the entrypoint launches the app under
+    ``opentelemetry-instrument`` (``OTEL_SDK_DISABLED=false``), and the exporter
+    target comes from environment variables fixed at container creation. This
+    sets those variables and force-recreates the ``lightspeed-stack`` container
+    so it exports HTTP/protobuf to the mock collector. ``after_feature`` reverts
+    these variables and recreates the container for later scenarios.
+    """
+    os.environ["OTEL_SDK_DISABLED"] = "false"
```
radofuchs, line 124: these steps require the recreation of the container, so why not set them directly in the docker compose to enable this all the time? If anything it will make sure that the application is able to handle this as it should

tests/e2e/features/opentelemetry.feature
```diff
@@ -1,4 +1,4 @@
-@cfg_authorized @OTel @skip
+@cfg_authorized @OTel @skip-in-prow
```
radofuchs, line 1: do we have a task to get this working also in the konflux environment? If not create one as we need it to work there as well

radofuchs (changes requested):

tests/e2e/features/steps/opentelemetry.py
```diff
@@ -0,0 +1,156 @@
...
+    assert (
+        response.status_code == 200
+    ), f"Failed to reset mock OTEL collector: HTTP {response.status_code}"
+
+
+def _poll_collector_contains(marker: str) -> bool:
+    """Return True once the collector has buffered a payload containing ``marker``."""
+    url = f"{MOCK_OTEL_CONTROL_BASE}/received"
+    deadline = time.monotonic() + _DELIVERY_TIMEOUT_S
+    while True:
```
tisnik, line 86: something like `while time.monotonic() < deadline:` put the condition into the loop control statement

tests/e2e/features/steps/opentelemetry.py
radofuchs: there is absolutely no reason to create this extra step, it only makes things slower. The better approach, as I pointed in my comment below, is just to enable it also in the docker compose by adding this env var there.

tests/e2e/mock_otel_collector/server.py
```diff
@@ -0,0 +1,143 @@
...
+import threading
+from collections import deque
+from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
+from typing import Any
+
+# Resource bounds so a reachable caller cannot exhaust process memory by sending
+# large or repeated exports (the Compose service publishes port 4318). The deque
+# caps the entry count (evicting oldest); the per-body cap bounds each entry.
+_MAX_BODY_BYTES = 5 * 1024 * 1024  # reject a single OTLP body larger than 5 MiB
+_MAX_ENTRIES = 1000  # keep at most this many buffered exports
```
tisnik, line 39: 5GB in total?

tests/e2e/mock_otel_collector/server.py
```diff
@@ -0,0 +1,143 @@
...
+    def _handle_received_query(self, query: str) -> None:
+        """Answer a count or substring query over the buffered exports."""
+        contains = None
+        for pair in query.split("&"):
+            key, sep, value = pair.partition("=")
+            if sep and key == "contains":
+                contains = value
+                break
+
+        if contains:
```
tisnik, line 80: I'd use `contains is not None` if we really need to detect empty parameter

tests/e2e/mock_otel_collector/server.py
```diff
@@ -0,0 +1,143 @@
...
+            needle = contains.encode("utf-8")
+            with _lock:
+                matches = sum(1 for body in _received if needle in body)
```
tisnik, line 83: can be replaced by len(for ...), but whatever

tests/e2e/mock_otel_collector/server.py
```diff
@@ -0,0 +1,143 @@
...
+        try:
+            length = int(raw_length)
+        except ValueError:
+            self._send_json(400, {"error": "invalid content-length"})
+            return
+        if length < 0 or length > _MAX_BODY_BYTES:
+            self._send_json(413, {"error": "payload too large"})
+            return
+
+        body = self.rfile.read(length) if length else b""
```
tisnik, line 120: is it test for None? or zero?

tests/e2e/mock_otel_collector/server.py
anik120: Right, 1000 entries × 5 MiB was a ~5 GB worst case. Fixed by adding a total-bytes cap alongside the entry-count cap. The buffer now evicts oldest entries whenever either limit is exceeded [here](https://github.com/lightspeed-core/lightspeed-stack/pull/2609/changes#diff-c3db036ba30bccaacad9a8df69b04bc343d69adea57ce792e9de66331f7eb63fR62-R65). So the effective ceiling is now 64 MiB regardless of entry count. /reset zeroes both the deque and the running total.

tests/e2e/features/opentelemetry.feature
anik120: Do you mind helping me by creating that? I'm sure you know better than me where exactly the task should be created and tracked and what to write ;)

tests/e2e/mock_otel_collector/server.py
anik120: It's testing for zero, not None. length comes from int(raw_length) and negatives/oversized values are already rejected above, so it's always a non-negative int — never None. I've made that explicit by changing if length to if length > 0 (and clarified the comment), so a zero-length body just becomes b"" without a read.

anik120:
@radofuchs @tisnik PTAL. Pivoted the whole thing to always-on OTEL export: the OTEL_* vars now default to export-on pointing at `mock-otel` directly in `docker-compose.yaml`/`docker-compose-library.yaml`, and lightspeed-stack lists `mock-otel` under `depends_on`. That removes the need to force-recreate the container per scenario, so I've deleted `force_recreate_compose_service` and `compose_file_for_mode` altogether. The step definition is now simply: 

"Wait for the collector's Docker health + reset its buffer."

tests/e2e/features/opentelemetry.feature
radofuchs: created https://redhat.atlassian.net/browse/LCORE-4183

radofuchs (approved):
LGTM

tisnik (approved):
LGTM

PR #2611, opened 2026-09-04
LCORE-1673: e2e feature file for conversation compaction (no step implementation)

max-svistunov:
## Description

Implements LCORE-1673: the behave feature file for conversation compaction, authored from `docs/design/conversation-compaction/conversation-compaction.md` ahead of the step definitions (LCORE-2230, #2612), so the scenarios describe intended behaviour rather than the implementation.

> **Since the last review round:** reworked per the changes-requested review. Six scenarios became three, the threshold crossing is deterministic, the fixtures are provider-agnostic and shield-free, the feature is `@skip` until #2612 lands the steps, and the branch is rebased onto current `main` (OGX rename applied to the server-mode fixtures).
>
> **Since the second review round:** the query scenario gains queries 4 and 5, which trigger a second compaction and check that the first summary survives it (details below). Buffered turns are not checked past query 3, because they are dropped from later requests after a compaction; that is a production bug, tracked as LCORE-4219.

The scenarios observe compaction strictly from outside the deployed stack, per the test-layer boundary in `docs/testing/e2e_testing.md`:

| Scenario | What it shows |
|---|---|
| the third query crosses the threshold, a second summary keeps the first, recall and history survive | `context_status` is `full` on turns 1 and 2 and `summarized` on turns 3-5; the turn-3 answer recalls one fact from the summarized turn and one from the verbatim buffer; turn 5 creates a second summary and its answer recalls a fact only the first summary holds; GET conversation still returns the earlier user queries |
| the native stream announces compaction on the query that crosses the threshold | the `compaction` SSE event precedes the first token on turn 3; the `end` event reports `full` on turns 1 and 2 and `summarized` on turn 3 |
| compaction stays off when disabled, even past the threshold | same three turns with `enabled: false` keep `context_status` at `full` |

**Why the crossing is deterministic.** Turns 1 and 2 each plant one fact and ask for "OK" only. Turn 3 is a fixed 248-token query, above the 200-token threshold (2000-token window × 0.1, 100-token floor) on its own. With `buffer_turns: 1` nothing can be summarized before turn 3, so `full` on turns 1 and 2 does not depend on response length, and `summarized` on turn 3 does not either. Turn 3 also asks for both facts, so one answer covers summary recall and buffer recall.

**Why the second compaction is on turn 5.** After turn 3 only turn 3 follows the summary marker, and it fits the one-turn buffer, so turn 4 (which plants a database name) summarizes nothing. On turn 5, turns 3 and 4 follow the marker: turn 3 goes into a second summary and turn 4 stays verbatim. Turn 5 asks for the datacenter name, which turn 1 planted and which exists only in the first summary: turn 3 asks for the cluster and namespace names, so its answer never repeats the datacenter name, and the second summary cannot supply it. If the second summary replaced the first, turn 5 could not answer. This was checked by running the scenario's exact queries through the real `apply_compaction` with the conversation store and the summarization call mocked.

**Fixtures:** `lightspeed-stack-compaction.yaml` and `lightspeed-stack-compaction-disabled.yaml` per mode. `inference.context_windows` lists every provider/model pair the e2e workflows run (openai, azure, google-vertex, watsonx, aws-bedrock) at 2000 tokens; the real windows are untouched, this only drives the local estimate. Mapping keys are not env-substituted, so the vLLM runs (rhaiis, rhelai, model id from an env var) cannot be listed; #2612 adds a gate step that skips the scenarios when the active model has no registered window. The `pii-redaction` shield from the base fixture is dropped: it cost tokens on every turn and its `\d+` rule would have rewritten the cluster name the recall assertion looks for.

**Step definitions are out of scope here.** Nine step patterns are undefined by design and the feature carries a feature-level `@skip` (Konflux runs the whole test list); #2612 implements the steps, removes the tag and wires `@cfg_compaction` into the CI shards.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [x] E2E tests improvement
- [ ] Other (please describe):

## Tools used to create PR

- Related Issue # LCORE-1673
- Closes # LCORE-1673

## Testing

**Parse check (ticket acceptance criterion):**
```
uv run behave --dry-run -f plain tests/e2e/features/conversation-compaction.feature
3 scenarios; 61 steps, 11 undefined (the patterns for LCORE-2230)
```

**Whole suite still parses with the entry in `test_list.txt`, and the tag expression the shards use excludes the feature while it is `@skip`:**
```
uv run behave --dry-run -f plain @tests/e2e/test_list.txt          # 399 scenarios, no parse errors
uv run behave --dry-run --tags="not @skip and @cfg_compaction" ...  # 54 skipped, 0 untested
```

**Fixtures load through the real `Configuration` model** (both modes, both variants): `compaction.enabled` true/false, `threshold_ratio` 0.1, `token_floor` 100, `buffer_turns` 1, five `context_windows` entries, zero shields.

**Token arithmetic:** the turn-3 query is 248 tokens under cl100k_base; the system prompt the e2e stack uses is 5 tokens; threshold is 200.

## Summary by CodeRabbit

- **Tests**
  - Added end-to-end configurations for conversation compaction in library and server modes.
  - Added scenarios covering threshold-triggered compaction, disabled compaction, streaming conversations, transcript handling, retrieval, and sensitive-number redaction.
  - Registered the conversation-compaction feature in the end-to-end test suite.
  - Consolidated the scenarios and temporarily skipped execution until the required step definitions are available.

### DIFF
--- a/tests/e2e/configuration/library-mode/lightspeed-stack-compaction-disabled.yaml
+++ b/tests/e2e/configuration/library-mode/lightspeed-stack-compaction-disabled.yaml
@@ -1,0 +1,62 @@
+name: Lightspeed Core Service (LCS)
+service:
+  host: 0.0.0.0
+  port: 8080
+  auth_enabled: false
+  workers: 1
+  color_log: true
+  access_log: true
+ogx:
+  # Library mode - embeds OGX as library
+  use_as_library_client: true
+  # Unified mode: run.yaml (materialized per provider by CI/the harness)
+  # is consumed as the synthesis profile instead of the legacy two-file path.
+  config:
+    profile: run.yaml
+user_data_collection:
+  feedback_enabled: true
+  feedback_storage: "/tmp/data/feedback"
+  transcripts_enabled: true
+  transcripts_storage: "/tmp/data/transcripts"
+authentication:
+  module: "noop"
+inference:
+  default_provider: openai
+  default_model: gpt-4o-mini
+  # Compaction e2e (LCORE-1673): a deliberately small window for the CI
+  # model so a three-turn conversation crosses the trigger threshold.
+  # The real model window is far larger; this only drives the local
+  # estimate, never the provider.
+  context_windows:
+    openai/gpt-4o-mini: 2000
+rag:
+  byok:
+    stores:
+      - rag_id: e2e-test-docs
+        backend: faiss
+        embedding_model: sentence-transformers/all-mpnet-base-v2
+        embedding_dimension: 768
+        vector_db_id: ${env.FAISS_VECTOR_STORE_ID}
+        db_path: ${env.KV_RAG_PATH:=~/.llama/storage/rag/kv_store.db}
+        score_multiplier: 1.0
+  retrieval:
+    tool:
+      sources:
+        - e2e-test-docs
+
+shields:
+  - name: pii-redaction
+    provider_id: redaction
+    config:
+      rules:
+        - pattern: '\d+'
+          replacement: '[NUM]'
+
+# Same small window and threshold as lightspeed-stack-compaction.yaml, but
+# compaction switched off: context_status must stay "full" past the
+# threshold (enabled is a full off-switch).
+compaction:
+  enabled: false
+  threshold_ratio: 0.1
+  token_floor: 100
+  buffer_turns: 1
--- a/tests/e2e/configuration/library-mode/lightspeed-stack-compaction-disabled.yaml
+++ b/tests/e2e/configuration/library-mode/lightspeed-stack-compaction-disabled.yaml
@@ -23,12 +23,19 @@ authentication:
 inference:
   default_provider: openai
   default_model: gpt-4o-mini
-  # Compaction e2e (LCORE-1673): a deliberately small window for the CI
-  # model so a three-turn conversation crosses the trigger threshold.
-  # The real model window is far larger; this only drives the local
-  # estimate, never the provider.
+  # Compaction e2e (LCORE-1673): a deliberately small window for every
+  # model the e2e workflows run against, so the third query of the
+  # compaction scenarios crosses the trigger threshold. The real windows
+  # are far larger; this only drives the local estimate, never the provider.
+  # The vLLM-backed runs (rhaiis, rhelai) take the model id from an env
+  # var, and context_windows keys are not env-substituted, so they are
+  # not listed and skip the token-based trigger.
   context_windows:
     openai/gpt-4o-mini: 2000
+    azure/gpt-4o-mini: 2000
+    google-vertex/publishers/google/models/gemini-2.5-flash: 2000
+    watsonx/meta-llama/llama-3-3-70b-instruct: 2000
+    aws-bedrock/deepseek.v3-v1:0: 2000
 rag:
   byok:
     stores:
@@ -43,14 +50,6 @@ rag:
     tool:
       sources:
         - e2e-test-docs
-
-shields:
-  - name: pii-redaction
-    provider_id: redaction
-    config:
-      rules:
-        - pattern: '\d+'
-          replacement: '[NUM]'
 
 # Same small window and threshold as lightspeed-stack-compaction.yaml, but
 # compaction switched off: context_status must stay "full" past the
--- a/tests/e2e/configuration/library-mode/lightspeed-stack-compaction.yaml
+++ b/tests/e2e/configuration/library-mode/lightspeed-stack-compaction.yaml
@@ -1,0 +1,61 @@
+name: Lightspeed Core Service (LCS)
+service:
+  host: 0.0.0.0
+  port: 8080
+  auth_enabled: false
+  workers: 1
+  color_log: true
+  access_log: true
+ogx:
+  # Library mode - embeds OGX as library
+  use_as_library_client: true
+  # Unified mode: run.yaml (materialized per provider by CI/the harness)
+  # is consumed as the synthesis profile instead of the legacy two-file path.
+  config:
+    profile: run.yaml
+user_data_collection:
+  feedback_enabled: true
+  feedback_storage: "/tmp/data/feedback"
+  transcripts_enabled: true
+  transcripts_storage: "/tmp/data/transcripts"
+authentication:
+  module: "noop"
+inference:
+  default_provider: openai
+  default_model: gpt-4o-mini
+  # Compaction e2e (LCORE-1673): a deliberately small window for the CI
+  # model so a three-turn conversation crosses the trigger threshold.
+  # The real model window is far larger; this only drives the local
+  # estimate, never the provider.
+  context_windows:
+    openai/gpt-4o-mini: 2000
+rag:
+  byok:
+    stores:
+      - rag_id: e2e-test-docs
+        backend: faiss
+        embedding_model: sentence-transformers/all-mpnet-base-v2
+        embedding_dimension: 768
+        vector_db_id: ${env.FAISS_VECTOR_STORE_ID}
+        db_path: ${env.KV_RAG_PATH:=~/.llama/storage/rag/kv_store.db}
+        score_multiplier: 1.0
+  retrieval:
+    tool:
+      sources:
+        - e2e-test-docs
+
+shields:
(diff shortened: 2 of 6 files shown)


tests/e2e/configuration/library-mode/lightspeed-stack-compaction-disabled.yaml
```diff
@@ -0,0 +1,62 @@
...
+    profile: run.yaml
+user_data_collection:
+  feedback_enabled: true
+  feedback_storage: "/tmp/data/feedback"
+  transcripts_enabled: true
+  transcripts_storage: "/tmp/data/transcripts"
+authentication:
+  module: "noop"
+inference:
+  default_provider: openai
```
radofuchs, line 24: is there any specific reason why we hardcode here the specific model and ignore the other providers? I would rather see here the env var for the model and provider and specific context windows for all applicable providers

tests/e2e/configuration/library-mode/lightspeed-stack-compaction-disabled.yaml
```diff
@@ -0,0 +1,62 @@
...
+        vector_db_id: ${env.FAISS_VECTOR_STORE_ID}
+        db_path: ${env.KV_RAG_PATH:=~/.llama/storage/rag/kv_store.db}
+        score_multiplier: 1.0
+  retrieval:
+    tool:
+      sources:
+        - e2e-test-docs
+
+shields:
+  - name: pii-redaction
```
radofuchs: adding shields into this config does not seem like a good idea, given the fact that it consumes even more tokens

tests/e2e/features/conversation-compaction.feature
```diff
@@ -0,0 +1,165 @@
...
+  three-turn conversation crosses it.
+
+  Background:
+    Given The service is started locally
+      And The system is in default state
+      And REST API service prefix is /v1
+      And the Lightspeed stack configuration directory is "tests/e2e/configuration"
+
+
+  Scenario: context_status reports full while compaction never triggers
```
radofuchs: I am completely missing the idea behind this case, there is no difference between this one and the disabled compaction one

tests/e2e/features/conversation-compaction.feature
```diff
@@ -0,0 +1,165 @@
...
+     When I use "query" to ask question
+     """
+     {"query": "Say hello", "model": "{MODEL}", "provider": "{PROVIDER}"}
+     """
+     Then The status code of the response is 200
+      And The response context_status is "full"
+
+
+  @openai-only
+  Scenario: context_status reports summarized once the conversation crosses the threshold
```
radofuchs: how do we know in this test that the summarized expected output was actually set only when we reached the threshold and not earlier/later? Adding three queries which could cause the threshold being crossed at any point do not seem like a good idea at all, since you have no idea how many tokens will get consumed during the process. I would rather want to see here that the context_status gets updated to summarized after the specific prompts crossing that limit, meaning that I want you to assert that the response context_status is full before the query and summarized after it

tests/e2e/features/conversation-compaction.feature
```diff
@@ -0,0 +1,165 @@
...
+     When I use "query" to ask question with same conversation_id
+     """
+     {"query": "Explain what a deployment is in about five sentences.", "model": "{MODEL}", "provider": "{PROVIDER}"}
+     """
+     Then The status code of the response is 200
+      And The response context_status is "summarized"
+
+
+  @openai-only
+  Scenario: the assistant still recalls what was said before the summary
```
radofuchs: this test completely covers the logic of the previous one, so the previous one can be removed

tests/e2e/features/conversation-compaction.feature
```diff
@@ -0,0 +1,165 @@
...
+     """
+     Then The status code of the response is 200
+      And The response context_status is "summarized"
+     When I use "query" to ask question with same conversation_id
+     """
+     {"query": "What is the name of my cluster? Reply with the name only.", "model": "{MODEL}", "provider": "{PROVIDER}"}
+     """
+     Then The status code of the response is 200
+      And The response context_status is "summarized"
+      And The response contains following fragments
```
radofuchs, line 43: is this intended to cover the logic using the buffer? If so, then I also want to see the test here to test that something is not rememebered

tests/e2e/features/conversation-compaction.feature
```diff
@@ -0,0 +1,165 @@
...
+          | Fragments in LLM response |
+          | aurora-prod-7             |
+
+
+  @openai-only
+  Scenario: the full conversation history stays available after compaction
```
radofuchs: this case can be merged with the previous one exercising the same logic

tests/e2e/features/conversation-compaction.feature
```diff
@@ -0,0 +1,165 @@
+@cfg_compaction
```
radofuchs: skip this feature, it will cause konflux test run to fail like this

tests/e2e/features/conversation-compaction.feature
```diff
@@ -0,0 +1,165 @@
...
+Feature: Conversation compaction
+
+  When a conversation's estimated input approaches the model's context
```
radofuchs: this description is very long, remove the R references as no one will trace it back to the design doc later and these are causing only confusion

radofuchs (changes requested):

tests/e2e/features/conversation-compaction.feature
max-svistunov: Partly -- one fact summarized, one buffered. What would "not remembered" look like here?

max-svistunov:
@radofuchs Could you PTAL again? All comments addressed, one question for you inline.

tests/e2e/features/conversation-compaction.feature
snuryyeva: It looks like summarization is additive, meaning when we summarize for the first time, it will remember it for the rest of the conversation unless we hit some sort of the limit in summary context. At that point it has to re-summarize and this is when we could see NOT remembered case.

snuryyeva: That being said, I think it is worth to add one/two more rounds of query to this scenario here to trigger second cycle of compaction to verify that when second summary is added, we still retain the compaction details from the first summary ( basically check that R3 - that second compaction is NOT overwriting the first compaction cycle)  
Does this make sense ? @max-svistunov , @radofuchs

snuryyeva (changes requested):

tests/e2e/features/conversation-compaction.feature
radofuchs: the buffered part is nucely covered in the integration tests, so that can be left as is here. But I agree with @snuryyeva about the second compaction

max-svistunov: @snuryyeva @radofuchs Thanks, done.

tests/e2e/configuration/library-mode/lightspeed-stack-compaction.yaml
```diff
@@ -0,0 +1,60 @@
...
+inference:
+  default_provider: openai
+  default_model: gpt-4o-mini
+  # Compaction e2e (LCORE-1673): a deliberately small window for every
+  # model the e2e workflows run against, so the third query of the
+  # compaction scenarios crosses the trigger threshold. The real windows
+  # are far larger; this only drives the local estimate, never the provider.
+  # The vLLM-backed runs (rhaiis, rhelai) take the model id from an env
+  # var, and context_windows keys are not env-substituted, so they are
+  # not listed and skip the token-based trigger.
```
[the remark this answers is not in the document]
max-svistunov: Handled in #2612 -- the scenarios get skipped when the active model has no context_windows entry, so the vLLM runs skip them.

radofuchs (approved):
LGTM

snuryyeva (approved):
LGTM

PR #2612, opened 2026-09-04
LCORE-2230: behave step definitions for the conversation-compaction feature

max-svistunov:
## Description

Implements LCORE-2230: the step definitions for `conversation-compaction.feature` (LCORE-1673, #2611, now merged), plus the CI wiring. After this lands the three compaction scenarios go from skipped to executing.

Five patterns in `tests/e2e/features/steps/conversation_compaction.py`, all observing the deployed stack from outside per the test-layer boundary in `docs/testing/e2e_testing.md`:

| Step | Reads |
|---|---|
| `the active model has a registered context window` | the fixture file the last `The service uses ...` step applied; fails when `context.default_provider/default_model` has no `inference.context_windows` entry |
| `The response context_status is "<status>"` | `/v1/query` JSON body |
| `The conversation history includes the following user queries` | every `user` message across `chat_history` from GET conversation |
| `The streamed response contains a compaction event before the first token` | SSE `data:` payloads in order; `compaction` must precede the first `token` |
| `The streamed response end event has context_status "<status>"` | the `end` event payload |

**Why the context-window step, and why the vLLM tag.** The provider and rhaiis workflows run `make test-e2e`, the whole test list with only `@skip` excluded. The vLLM-backed runs take their model id from an env var, and `context_windows` keys are not env-substituted, so those runs cannot have a window entry, the trigger never fires and a `summarized` assertion could never pass. Per review, a missing entry now **fails** the scenario instead of skipping it: on every other matrix the provider/model pair is fixed and the fixtures list it (openai, azure, google-vertex, watsonx, aws-bedrock), so a missing entry means a fixture and a workflow disagree. The vLLM matrices are excluded one level up instead, by a feature-level `@skip-on-vllm` tag handled in `before_scenario` next to the existing `@openai-only` check, so they never reach the step. The step resolves the fixture *source* path the same way `configure_service` does, not the repo-root copy, which Prow never writes because the config goes into a ConfigMap.

**CI:** the feature loses its `@skip` tag; `@cfg_compaction` joins the `other` shard in `e2e_tests.yaml` and the Makefile's default `E2E_BEHAVE_TAG_EXPR`, so the feature runs in both deployment modes.

SSE decoding is local to the module: the shared `_parse_streaming_response` keeps only what existing steps use and drops unknown events, and these assertions need event order, not a reconstructed answer.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [x] CI configuration change
- [ ] Unit tests improvement
- [x] E2E tests improvement
- [ ] Other (please describe):

## Tools used to create PR

- Related Issue # LCORE-2230
- Closes # LCORE-2230

## Testing

**Zero undefined steps (ticket acceptance criterion):**
```
uv run behave --dry-run -f plain tests/e2e/features/conversation-compaction.feature
3 scenarios, 64 steps, 0 undefined
```

**Lint on the new module:** black and ruff clean; pylint reports only the `not-callable` / `import-error` noise every behave step module in this repo gets.

**Gate resolver:** checked by hand for both modes; resolves to `tests/e2e/configuration/<mode>/lightspeed-stack-compaction.yaml`.

**Execution, and a caveat.** The GitHub e2e workflow runs on `pull_request_target`, so its shard tag expressions come from `main`, not from this PR. The `other` shard on this PR therefore still excludes `@cfg_compaction` (checked in the job log: `E2E_BEHAVE_TAG_EXPR=not @skip and (@cfg_rh_identity or @cfg_negative or @cfg_byok_pdf or @cfg_degraded or @cfg_unified)`), and the three scenarios do not run in PR CI until this change is on `main`. Konflux runs the whole test list and does exercise them. On an earlier revision it ran them live on OpenShift against `openai/gpt-4o-mini`: the query scenario and the disabled scenario passed, including `full` on turns 1-2, `summarized` on turn 3, recall of both facts and the conversation history, which confirms the threshold arithmetic (2000-token window, 10% ratio, 100-token floor, 248-token third query). The streaming scenario errored in the gate step: it read the config basename from the context, which behave drops at the end of a scenario, and `configure_service` does not set it again when the next scenario asks for the same YAML. Fixed by reading `get_active_lightspeed_stack_config_basename()` instead; the next Konflux run passed. The branch has since been rebuilt on #2611's second-compaction change (queries 4 and 5 of the query scenario), without the empty CI re-trigger commits; the new queries need no new step patterns.

## Summary by CodeRabbit

- **Tests**
  - Added end-to-end coverage for conversation compaction in library and server modes.
  - Verified context status transitions, preservation of conversation history, streaming compaction events, and behavior when compaction is disabled.
  - Included configurations covering retrieval, redaction, authentication, and supported inference models.
  - Added the new compaction scenarios to the standard end-to-end test suite.

- **Documentation**
  - Documented the new conversation compaction test steps.

### DIFF
--- a/tests/e2e/configuration/library-mode/lightspeed-stack-compaction-disabled.yaml
+++ b/tests/e2e/configuration/library-mode/lightspeed-stack-compaction-disabled.yaml
@@ -1,0 +1,62 @@
+name: Lightspeed Core Service (LCS)
+service:
+  host: 0.0.0.0
+  port: 8080
+  auth_enabled: false
+  workers: 1
+  color_log: true
+  access_log: true
+ogx:
+  # Library mode - embeds OGX as library
+  use_as_library_client: true
+  # Unified mode: run.yaml (materialized per provider by CI/the harness)
+  # is consumed as the synthesis profile instead of the legacy two-file path.
+  config:
+    profile: run.yaml
+user_data_collection:
+  feedback_enabled: true
+  feedback_storage: "/tmp/data/feedback"
+  transcripts_enabled: true
+  transcripts_storage: "/tmp/data/transcripts"
+authentication:
+  module: "noop"
+inference:
+  default_provider: openai
+  default_model: gpt-4o-mini
+  # Compaction e2e (LCORE-1673): a deliberately small window for the CI
+  # model so a three-turn conversation crosses the trigger threshold.
+  # The real model window is far larger; this only drives the local
+  # estimate, never the provider.
+  context_windows:
+    openai/gpt-4o-mini: 2000
+rag:
+  byok:
+    stores:
+      - rag_id: e2e-test-docs
+        backend: faiss
+        embedding_model: sentence-transformers/all-mpnet-base-v2
+        embedding_dimension: 768
+        vector_db_id: ${env.FAISS_VECTOR_STORE_ID}
+        db_path: ${env.KV_RAG_PATH:=~/.llama/storage/rag/kv_store.db}
+        score_multiplier: 1.0
+  retrieval:
+    tool:
+      sources:
+        - e2e-test-docs
+
+shields:
+  - name: pii-redaction
+    provider_id: redaction
+    config:
+      rules:
+        - pattern: '\d+'
+          replacement: '[NUM]'
+
+# Same small window and threshold as lightspeed-stack-compaction.yaml, but
+# compaction switched off: context_status must stay "full" past the
+# threshold (enabled is a full off-switch).
+compaction:
+  enabled: false
+  threshold_ratio: 0.1
+  token_floor: 100
+  buffer_turns: 1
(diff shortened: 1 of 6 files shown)


tests/e2e/features/steps/conversation_compaction.py
```diff
@@ -0,0 +1,166 @@
...
+    provider/model pair comes from the context, which already reflects the
+    ``E2E_DEFAULT_*_OVERRIDE`` values on every path.
+    """
+    fixture_path = _active_fixture_path(context)
+    with open(fixture_path, encoding="utf-8") as config_file:
+        config = yaml.safe_load(config_file) or {}
+    windows = (config.get("inference") or {}).get("context_windows") or {}
+    model_key = f"{context.default_provider}/{context.default_model}"
+    if model_key not in windows:
+        context.scenario.skip(
```
radofuchs: this should not skip the scenario but fail it, since if we do not have it set but still use this step then this is a problem, not something to be ignored

radofuchs (changes requested):

radofuchs (approved):
LGTM

PR #2614, opened 2026-09-04
LCORE-3963: render the full ADF document in fetch-jira output, and fix its error handling

max-svistunov:
## Description

`dev-tools/fetch-jira.sh` renders a Jira ticket's ADF description and comments to text. Its extractor handled only strong/code marks, lists, headings and code blocks, so most of what a ticket actually contains was dropped or flattened. Link targets vanished, leaving bare text like "PR #2451" with nothing to follow. Smart links produced nothing at all, even though the URL is their entire content. Mentions, tables, attachments, status lozenges, dates, task checkboxes, panel types and ordered-list numbering all disappeared or lost their meaning. Since most bug reports reference other tickets and PRs by smart link, the tool was hiding exactly the references a reader needs.

This PR renders all of it, and fixes several problems found while doing so.

**ADF coverage.** Link marks render as `text <href>`, with Jira's `#icft=` click-tracking fragment stripped. `inlineCard`, `blockCard` and `embedCard` all resolve their URL, from either `attrs.url` or the resolved `attrs.data.url`. Tables become markdown with a rule under header rows. Also handled: mentions, `rule`, blockquotes, panels (keeping their type), `expand`/`nestedExpand` (keeping the title that says what is collapsed), task items (keeping the checkbox), decisions, attachments, emoji, dates, status lozenges, the `em` and `strike` marks, the `codeBlock` language, and ordered-list numbering.

**Paragraph rendering.** The extractor returns one entry per line and the block walk extended that list for every child, so a paragraph of inline nodes came out one node per line — a sentence containing a link was chopped into fragments. Paragraphs now join their inline children into one line, with `hardBreak` the only thing that splits them.

**Error handling.** The script runs under `set -euo pipefail`, which defeated its own error handling. The main issue fetch assigned from an unguarded `curl`, so a connection failure aborted before the `Error fetching` branch below it could run; that branch was dead for anything but an HTTP error with a JSON body. The two JQL lookups piped `curl` into `python3`, where `pipefail` turned a curl failure into a failed assignment. `fetch_ticket` then returns 1 after reporting a failure and every call site was unguarded, so the first unreachable ticket ended the whole run. Every requested ticket is now attempted and reported, and the script still exits non-zero if any failed.

**Silent truncation.** The comment endpoint pages at 100 and the child search asks for 20, and a partial result was presented exactly like a complete one. Neither is worth paginating here, but both responses say when they are partial, so that is now reported.

**Quoting.** All four embedded Python programs were passed inline to `python3 -c`, which puts them inside a double-quoted shell word — every quote, dollar sign and backtick in ~200 lines of Python was shell syntax first and Python second. That is not hypothetical: a comment containing the word `"here"` in double quotes closed the shell string and reopened it, which shellcheck reported as SC2140 and CI failed on. Each program moves into a quoted heredoc, so the Python is written as Python. This commit is a pure move, verified by diffing the output of four invocations before and after.

Closes LCORE-3963.

## Type of change

- [x] Other (please describe): developer tooling (dev-tools)

## Tools used to create PR

- Related Issue # LCORE-3963
- Closes # LCORE-3963

## Testing

Requires `~/.config/jira/credentials.json`. No unit tests: this is a standalone dev script with no test harness in the repo.

1. Check the linter CI job passes locally, since an earlier revision of this branch failed it:

   ```
   shellcheck -- */*.sh && bash -n dev-tools/fetch-jira.sh
   ```

   Both are clean. `make shellcheck` runs the same command in CI.

2. Fetch a ticket whose description is link-heavy and confirm the targets appear:

   ```
   dev-tools/fetch-jira.sh 3788
   ```

   LCORE-3788's five smart links now render as `<https://redhat.atlassian.net/browse/LCORE-1572>` and similar; before this branch, `grep -c '<https' ` on that output returned 0. The tracking fragment (`#icft=LCORE-1572`) is stripped. Its two `orderedList` nodes now number `1.`–`2.` and `1.`–`4.` where they previously rendered as bullets; its two `bulletList` nodes are unchanged. Prose that was previously one fragment per line now reads as sentences.

3. Confirm comments and recursion still render:

   ```
   dev-tools/fetch-jira.sh --comments 3883
   dev-tools/fetch-jira.sh --linked-depth 1 3788
   ```

   Code blocks and lists in the comment thread are byte-identical to the output before this branch.

4. Confirm a failure no longer ends the run. Point `instance` in the credentials file at a closed port, then:

   ```
   dev-tools/fetch-jira.sh 3963 3883
   ```

   Both tickets are reported and the exit status is 1:

   ```
   Error fetching LCORE-3963

   ────────────────────────────────────────────────────────

   Error fetching LCORE-3883
   ```

   Before this branch the same command printed nothing and exited 7, because `set -e` killed the script inside the command substitution.

5. The extractor was also driven directly against a synthetic ADF document covering every node type and mark handled here, including cases the live tickets do not contain: multi-paragraph blockquotes and table cells, an `inlineCard` carrying `attrs.data.url` instead of `attrs.url`, a link mark combined with a `code` mark in both orders, an autolinked bare URL, and both truncation notices in the firing and non-firing directions.

## Summary by CodeRabbit

* **Improvements**
  * Jira descriptions and comments now support richer formatting, including links, mentions, lists, tables, code blocks, panels, attachments, and other structured content.
  * Jira links are cleaned to remove tracking fragments.
  * Extracted content better preserves indentation and formatting.
  * Comment output indicates when results were truncated after 100 comments.
  * Multi-ticket and recursive fetches continue processing when individual requests fail, while clearly reporting an overall failure status.

### DIFF
--- a/dev-tools/fetch-jira.sh
+++ b/dev-tools/fetch-jira.sh
@@ -147,9 +147,22 @@ def extract_text(node, depth=0):
                     text = f'**{text}**'
                 elif m.get('type') == 'code':
                     text = f'\`{text}\`'
+                elif m.get('type') == 'link':
+                    # Keep the target: a link whose text differs from its
+                    # href (ticket keys, "here", PR titles) is otherwise lost.
+                    href = m.get('attrs', {}).get('href', '')
+                    if href and href != text:
+                        text = text + ' <' + href + '>'
             return [text]
+        if ntype == 'inlineCard':
+            # Smart links (pasted Jira/GitHub URLs) carry the URL only here.
+            return ['<' + node.get('attrs', {}).get('url', '') + '>']
+        if ntype == 'mention':
+            return [node.get('attrs', {}).get('text', '@?')]
         if ntype == 'hardBreak':
             return ['\n']
+        if ntype == 'rule':
+            return ['---']
         if ntype == 'listItem':
             child_text = []
             for c in node.get('content', []):
@@ -170,6 +183,22 @@ def extract_text(node, depth=0):
             for c in node.get('content', []):
                 child_text.extend(extract_text(c, depth))
             return ['\`\`\`\n' + ''.join(child_text) + '\n\`\`\`']
+        if ntype == 'blockquote':
+            child_text = []
+            for c in node.get('content', []):
+                child_text.extend(extract_text(c, depth))
+            return ['> ' + l for l in ''.join(child_text).strip().split('\n')]
+        if ntype == 'table':
+            rows = []
+            for row in node.get('content', []):
+                cells = []
+                for cell in row.get('content', []):
+                    cell_text = []
+                    for c in cell.get('content', []):
+                        cell_text.extend(extract_text(c, depth))
+                    cells.append(''.join(cell_text).strip())
+                rows.append(' | '.join(cells))
+            return rows
         for c in node.get('content', []):
             lines.extend(extract_text(c, depth))
         if ntype == 'paragraph' and lines:
--- a/dev-tools/fetch-jira.sh
+++ b/dev-tools/fetch-jira.sh
@@ -142,21 +142,30 @@ def extract_text(node, depth=0):
         if ntype == 'text':
             text = node.get('text', '')
             marks = node.get('marks', [])
+            href = ''
             for m in marks:
                 if m.get('type') == 'strong':
                     text = f'**{text}**'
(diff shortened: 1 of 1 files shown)


tisnik (approved):
LGTM

tisnik (approved):
LGTM

PR #2646, opened 2026-09-09
LCORE-3390: Input guardrail point on all query endpoints

Jazzcort:
## Description

These three commits sequentially implement, unit-test, and integration-test the Granite Guardian input guardrail, a feature that evaluates user input against configured risk categories using logprob scoring to short-circuit unsafe runs. The first commit builds the core capability and helper utilities in _capability.py and utils.py while wiring them into build_agent and build_shield. The second introduces comprehensive unit tests covering safe and unsafe pathways, logprob extraction, token accumulation, and edge cases. Finally, the third commit adds end-to-end integration tests covering both the standalone shield execution (/responses, /rlsapi /infer) and pydantic-ai agent execution (/query, /streaming_query).

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [ ] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

Claude opus 4.6 is used to generate drafts for the test. The refinement was done by me.

- Related Issue [LCORE-3390](https://redhat.atlassian.net/browse/LCORE-3390)
- Closes #

## Testing
1. Get the models.corp's access
2. Configure the guardrail in the config file (lightspeed-stack.yaml)
3. Host the lightspeed stack instance and test the guardrail through out the inference endpoints

## Summary by CodeRabbit

- **New Features**
  - Added Granite Guardian safety moderation for input, output, and tool guardrail points.
  - Blocked requests are rejected before agent processing, with moderation results and conversation persistence where available.
  - Added Granite Guardian support to shield configuration and moderation workflows.
  - Added reusable guardrail point definitions for API configuration.
  - Added standalone moderation results indicating whether content passed or was blocked.

- **Tests**
  - Added unit and integration coverage for moderation decisions, streaming, usage tracking, persistence, and endpoint behavior.


src/pydantic_ai_lightspeed/capabilities/granite_guardian/_capability.py
```diff
@@ -0,0 +1,256 @@
...
+)
+from pydantic_ai_lightspeed.capabilities.utils import (
+    extract_conversation_id,
+    message_to_str,
+)
+from utils.conversations import append_turn_to_conversation
+
+type Guardrail = tuple[str, str, float, str]
+
+MODEL_NAME = "ibm-granite/granite-guardian-4.1-8b"
```
max-svistunov: The Guardian model name is hard-coded to `ibm-granite/granite-guardian-4.1-8b`, and `GraniteGuardianConfig` has no field to override it. OpenAI-compatible servers route by model name, so a deployment serving the model under another name (for example an Ollama tag in local development) cannot use this capability. The config's `url` field description ("The model_id to use for the guard") also suggests the model was meant to be configurable.

If 4.1 is intentionally the only supported model -- the prompt block in `utils.py` is built for the 4.1 client-side format, and the design doc's benchmarked reference is 3.3-8B, which uses a different template -- that constraint should be stated in the config description and docs. Otherwise, suggest adding a `model` field to `GraniteGuardianConfig`, defaulting to the current value, and passing it to `OpenAIChatModel`.

max-svistunov (commented):
@Jazzcort Thanks! A few things from the ticket side:

* The ticket asks for input rules to run in parallel with per-rule latency logged. Right now `_run_risk_check` goes through the risks one by one (coderabbit caught that too) and there's no latency logging per risk. With a few risks configured that's a few guardian round trips on every request, so it's better to have this one in the PR.

* On /query and /streaming_query the guard runs inside the agent run, so a blocked input still goes through RAG retrieval first (the design doc says blocked input should skip RAG). Question validity works the same way today, so we can keep it as it is now in this PR -- I opened LCORE-4090 to move the input check before RAG on both endpoints.

* Btw the coderabbit comment about the httpx client is correct -- the design doc asks for each detector to have one long-lived client, what we do now is create one per request and never close it.

The inline comments are from Claude, they're smaller things, I went through them and they're valid.

src/pydantic_ai_lightspeed/capabilities/granite_guardian/_capability.py
```diff
@@ -0,0 +1,256 @@
...
+            ctx: The run context containing the user prompt and usage tracker.
+            handler: The handler to call if the input passes all guardrails.
+
+        Returns:
+            The agent run result, either a rejection or the handler's result.
+        """
+        user_prompt = message_to_str(ctx.prompt)
+
+        input_guardrails = _filter_guardrails(self.config.risks, GuardrailPoint.INPUT)
+        # TODO: We need to consider how we want to reveal the token usage for Granite Guardian,  # pylint: disable=fixme
```
max-svistunov, line 233: The Guardian token usage is discarded here, and `test_guardian_token_usage_is_not_included_in_user_token_usage` pins that behaviour. Keeping it out of the user's reported token counts is reasonable, but it is currently not recorded anywhere: each configured risk adds a Guardian call per request that neither quota, logs nor metrics account for. The design doc asks for per-rule outcomes and latencies in logs and metrics (R10). Suggest at least logging usage per risk in `_run_risk_check`, so operators can see what the guardrail costs. Compaction's summarization calls have the same gap (LCORE-3910), so a shared approach may make sense later.

tests/integration/endpoints/test_granite_guardian_integration.py
```diff
@@ -0,0 +1,821 @@
...
+    mock_holder.return_value.get_client.return_value = mock_client
+    return mock_client
+
+
+# ============================================================================
+# /query endpoint tests
+# ============================================================================
+
+
+class TestQueryGraniteGuardian:
```
max-svistunov, line 620: No test covers what happens when the Guardian endpoint fails -- unreachable, timed out, or returning a response without `logprobs` or without a parseable `<score>` tag. Today the standalone path in `run_shield_moderation_v2` maps the exception to an HTTP error via `map_agent_inference_error`, and the in-agent path raises out of the agent run (an HTTP error on /query, an `error` SSE event on /streaming_query). That is fail-closed, which matches the design doc's default posture, but the design doc also expects a blocked request to use the normal refusal shape with a distinct log line (R9). Suggest adding one test per path that pins the intended behaviour, so a later change cannot silently turn it into fail-open.

src/pydantic_ai_lightspeed/capabilities/granite_guardian/_capability.py
Jazzcort: addressed!

src/pydantic_ai_lightspeed/capabilities/granite_guardian/_capability.py
Jazzcort: This is the thing we need to address in a bigger scope. We need to figure a way out to surface the token usage for these LLM based shields. Basically, combining the token usage from different models does not make any sense and it's the reason why the token usage of Granite Guardian shield is not include into the user's chat model usage.

tests/integration/endpoints/test_granite_guardian_integration.py
Jazzcort: Added two tests to wrap_run to pin the fail-closed behavior:

  - test_risk_check_failure_propagates — UnexpectedModelBehavior from a malformed Guardian response bubbles up, handler never runs.
  - test_guardian_unreachable_propagates — ModelAPIError from an unreachable endpoint does the same.

  For the missing logprobs / unparseable <score> tag cases, those are already covered by existing unit tests on _run_risk_check and _extract_tokens_inside_score_tag — they all raise UnexpectedModelBehavior, and since wrap_run doesn't distinguish between reasons, the new propagation test covers them all. Adding per-scenario wrap_run tests would just be testing the same code path again.

max-svistunov (approved):
@Jazzcort LGTM, thanks! 👍

src/models/config.py
```diff
@@ -3217,7 +3224,19 @@ class GraniteGuardianConfig(ConfigurationBase):
     """Configuration for the Granite Guardian moderation guardrail."""
 
     url: str = Field(
-        ..., title="Base URL", description="The model_id to use for the guard"
+        ...,
+        title="Base URL",
+        description="Base URL of the OpenAI-compatible inference endpoint.",
+    )
+
+    model: str = Field(
```
jrobertboos: can this be changed to model_id for uniformity with question_validity?

src/models/config.py
```diff
@@ -3142,6 +3141,14 @@ class RedactionShieldConfiguration(ConfigurationBase):
     )
 
 
+class GuardrailPoint(StrEnum):
```
jrobertboos: I know we talked about this before but upon reviewing I think this should remain as a str literal as we don't need the full weight of an enum (especially because for other "type" config fields we use literals). @tisnik wdyt?

jrobertboos (changes requested):
Overall looks good but see the comments :)

src/models/config.py
```diff
@@ -3243,6 +3262,14 @@ class GraniteGuardianConfig(ConfigurationBase):
         ),
     )
 
+    batch_size: PositiveInt = Field(
```
jrobertboos: O also do you think that we should have something like batch size or just a `parallel: Bool`?

src/models/config.py
Jazzcort: That makes sense. Will change 😁

src/models/config.py
Jazzcort: Yeah, based on all the other "type" are literals, I've change my mind. Literals might be a better pick here.

src/models/config.py
Jazzcort: In Ask Red Hat's gitlab repo, they just send all the requests in parallel at once. My argument there is that it maximizes the parallelism but it does not consider the backend resource we have. 

1. Having too many parallel requests will hit rate limit. I encountered this while I was implementing the sliding window streaming output guardrail if we set the window to be too small and it was just with one risk.
2. Even though doing it in a batch style does not boost the performance as much as firing all request in parallel at once but it can potentially reduce cost. If the prompt violets the guardrail in the earlier batch then the later risk checks can be skipped.
3. Maximizing the parallelism can be achieve by setting batch_size >= len(risks)

Maybe `batch_size` is not a suitable name for it.

Perhaps we should do
```python
parallel: bool | int
```

when `parallel = True` we set the batch_size = len(risks) under the hood and for `parallel = False` we set it to 1

what do you think?

Jazzcort:
@jrobertboos 
Also, after digging through the pydantic-ai-harness, I'm still leaning toward using `wrap_run` for input guardrail. The reasons are:

1. `wrap_run` only fires once per agent run, but `wrap_model_request` fires on every model request (initial prompt, tool returns, retries) and that's why in their implementation it needs to check `ctx.run_step` to make sure it's the first request with user prompt.
3. With `wrap_run`, we can do a clean short-circuit. With `wrap_model_request` we provide the violation message with `SkipModelRequest` but it still need to go through all the output processing which means **the violation message will be check for output guardrail**. I think it's not quite right if we do that. 

I feel like it's much simpler to implement the input guardrail in `wrap_run`. We check the prompt which is the only thing we need to check and then we either short-circuit or call `handler()`

However, `after_output_processing` is a better choice for output guardrail in non-streaming path. And it also provide the `ModelRetry` option where we have an option to request another model request if the violation happens, even though we only block the response now. I'll use it instead of checking the `AgentRunResult` for the `await handler()` in `wrap_run`.

src/models/config.py
jrobertboos: I really like:
> `parallel: bool | int`

src/models/config.py
```diff
@@ -3217,7 +3219,19 @@ class GraniteGuardianConfig(ConfigurationBase):
...
 
     url: str = Field(
-        ..., title="Base URL", description="The model_id to use for the guard"
+        ...,
+        title="Base URL",
+        description="Base URL of the OpenAI-compatible inference endpoint.",
+    )
+
+    model_id: str = Field(
+        "ibm-granite/granite-guardian-4.1-8b",
```
asimurka, line 3226: What happens if this model is not configured?

src/pydantic_ai_lightspeed/capabilities/granite_guardian/_capability.py
```diff
@@ -0,0 +1,313 @@
...
+    Returns:
+        A tuple of (model_response, threshold, violation_message).
+    """
+    name, block, threshold, violation_message = guardrail
+    start = asyncio.get_event_loop().time()
+    result = await model_request(
+        model=model,
+        messages=[ModelRequest.user_text_prompt(prompt, instructions=block)],
+        model_settings=OpenAIChatModelSettings(
+            openai_logprobs=True, openai_top_logprobs=20
```
asimurka, line 69: Why do we use this setup?

src/pydantic_ai_lightspeed/capabilities/granite_guardian/_capability.py
```diff
@@ -0,0 +1,313 @@
...
+        run_moderation_guardrail_point: The guardrail point used by the
+            standalone ``run`` method.
+    """
+
+    config: GraniteGuardianConfig
+    run_moderation_guardrail_point: GuardrailPoint = "input"
+    _model: Model = field(init=False)
+    # Only one Granite Guardian shield should be configured; multiple entries are
+    # unsupported. A dict is used defensively so that a misconfiguration with two
+    # distinct configs does not cause one to silently overwrite the other's model.
```
asimurka: I'm not sure if this is something I would expect as a user. I would personally implement config validator that directly rejects multiple instances. But if this is aligned with feature design I'm ok with it.

src/pydantic_ai_lightspeed/capabilities/granite_guardian/_capability.py
```diff
@@ -0,0 +1,313 @@
...
+        http_client = httpx.AsyncClient(
+            verify=self.config.verify_ssl,
+            timeout=self.config.timeout,
+        )
+
+        # When we attach the API key to the request, we need to make sure we encrypt the
+        # request by communicating through https
+        base_url = httpx.URL(self.config.url)
+        if self.config.api_key is not None and base_url.scheme != "https":
+            raise ValueError(
```
asimurka: This should be also rejected by pydantic validator on config level not during runtime

asimurka (changes requested):

src/models/config.py
asimurka: Also, I noticed that we're not using OGX here. Is this correct way how to reference guardian models? Just asking...

jrobertboos: the goal of this shield is to satisfy ask rhs need. they use models.corp for granite guardian so we are just extending that pattern. I don't think we need to run it through llama stack when we can access it directly.

src/pydantic_ai_lightspeed/capabilities/granite_guardian/_capability.py
jrobertboos: this is from Ask RH implementation

src/pydantic_ai_lightspeed/capabilities/granite_guardian/_capability.py
jrobertboos: resolved

src/models/config.py
asimurka: Sure, my question was more like - is this correct way how to reference models? `<provider>/<model>` (just asking).

Jazzcort: @asimurka do you mean that if this can successfully receive 200 from the models.corp endpoints or you concern about the format? 🤔

src/pydantic_ai_lightspeed/capabilities/granite_guardian/_capability.py
Jazzcort: @jrobertboos have mentioned this to me. We will remove this limiter and just remove the comment here since the current implementation can work with the multiple instances already. My initial thought was an instance is already capable of running multiple risk checks so having another instance with many duplicate fields (i.e. url, model_id ... etc.) is redundant. I think we don't need to write some complex validation logic for the config just to strictly limit the user from setting multiple instances.

tisnik (changes requested):
Please rebase, so the e2e will pass.

src/models/config.py
asimurka: The question is very simple, if this model identification is proved. If yes, I'm fine

Jazzcort: It's proved. I've run this granite guardian input guardrail with lightspeed stack and it works well. 😁

Jazzcort:
@asimurka I think the comments are all addressed. Please approve this if there are no changes needed. 😁 

@tisnik Just rebased this patch on main

asimurka (approved):
LGTM

jrobertboos (approved):
LGTM

tisnik (approved):
LGTM

PR #2693, opened 2026-09-14
LCORE-4088: Fix MCP auth header "Bearer" duplication

jrobertboos:
## Description

Fixed a bug where MCP servers configured to receive the client's `Authorization` header (either via the `headers` allowlist, e.g. `headers: [Authorization]`, or via a resolved Kubernetes service-account token) would receive a duplicated `Authorization: Bearer Bearer <token>` header, causing downstream MCP servers to fail token decoding (e.g. `jwt.exceptions.DecodeError: Invalid header padding`).

Root cause: `get_mcp_tools()` and `apply_mcp_headers_to_explicit_tools()` in `src/utils/responses.py` pop the resolved `Authorization` header value (which already includes the `Bearer` scheme, since that's the correct on-the-wire HTTP header representation built by `build_mcp_headers()`) and assign it directly to OGX's `InputToolMCP.authorization` field. OGX's MCP tool runtime treats that field as a raw token/credential and unconditionally prepends its own `Bearer ` prefix before forwarding the request to the MCP server, so passing an already-prefixed value produced the duplicate scheme.

Fix: added `strip_bearer_prefix()` (and its idempotent counterpart `ensure_bearer_prefix()`) to `src/utils/mcp/mcp_headers.py`, and strip the scheme before assigning to OGX's `authorization` field in both `get_mcp_tools()` and `apply_mcp_headers_to_explicit_tools()`. Also refactored the equivalent (already-correct, but duplicated and case-sensitive) "add Bearer if missing" logic in `mcp_oauth_probe.py`'s `check_mcp_auth()` and `mcp_tools.py`'s `_prepare_mcp_request_headers()` to reuse the new `ensure_bearer_prefix()` helper, which also fixes a latent case-sensitivity bug (e.g. a lowercase `"bearer <token>"` value would previously have been double-prefixed to `"Bearer bearer <token>"`).

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] Bump-up service version
- [ ] CI configuration change
- [x] Unit tests improvement
- [ ] Benchmarks improvement

## Tools used to create PR

- Assisted-by: Cursor (Claude Sonnet 5)
- Generated by: N/A

- Related Issue #LCORE-4088
- Closes #LCORE-4088

## Testing

- Added/updated unit tests in `tests/unit/utils/mcp/test_mcp_headers.py` (`TestStripBearerPrefix`, `TestEnsureBearerPrefix`) and `tests/unit/utils/test_responses.py` covering:
  - `get_mcp_tools()` / `apply_mcp_headers_to_explicit_tools()` correctly strip the `Bearer` scheme when it comes from a propagated `Authorization` header (the exact reported scenario: `headers: [Authorization]` allowlist config, no `authorization_headers`) or a resolved Kubernetes token.
  - `ensure_bearer_prefix()` / `strip_bearer_prefix()` normalize scheme casing without double-prefixing.
- Manually verified against a real MCP server configured with the `headers: [Authorization]` allowlist to confirm the downstream server now receives a single, correctly-formed `Authorization: Bearer <token>` header instead of the duplicated scheme.
- `uv run make format` and `uv run make test-unit` (targeted: `tests/unit/utils/mcp/`, `tests/unit/utils/test_responses.py`, and affected endpoint tests) pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Authorization header handling for MCP connections.
  * Prevented duplicated `Bearer` prefixes when authentication headers are forwarded or configured.
  * Standardized exact-case `Bearer` prefix formatting for authentication values.
  * Preserved raw authorization tokens when passing values to MCP tools.
  * Added coverage for forwarded authentication headers and Bearer prefix handling.


src/utils/mcp/mcp_headers.py
```diff
@@ -129,6 +129,53 @@ def extract_propagated_headers(
     return propagated
 
 
+def strip_bearer_prefix(value: str) -> str:
```
tisnik, line 132: can be reduced to value.removeprefix("bearer ")

tisnik (changes requested):

src/utils/mcp/mcp_headers.py
jrobertboos: idk if this would work because `removeprefix` is case sensitive (and while I can't imagine why someone would use "bearer" instead of "Bearer" but we should probably support it), wdyt?

tisnik: that's interesting, because the spec requires "Bearer" https://datatracker.ietf.org/doc/html/rfc6750#section-2.1

and yes, there were some servers that reject "bearer" because of this

jrobertboos: ok so I will modify it so that it only supports "Bearer"

tisnik (approved):
LGTM

PR #2708, opened 2026-09-15
fix(OGX): Add pgvector to ogxlibdev deps for remote::pgvector vector_io

JslYoon:
## Description

Adds the `pgvector` Python client to the `ogxlibdev` dependency group so the OGX `remote::pgvector` vector_io provider can initialize.

Without it, starting the stack with a pgvector vector store crashes at provider init:

```
ModuleNotFoundError: No module named 'pgvector'
```

The provider imports `from pgvector.asyncpg import register_vector`, but the package was not declared alongside the existing `psycopg2-binary` / `asyncpg` drivers. This mirrors the other vector_io backend deps (`faiss-cpu` for inline::faiss) already listed under the same `# API vector_io` comment.

## Changes

- `pyproject.toml`: add `pgvector>=0.3.6` under the `ogxlibdev` group (vector_io section).
- `uv.lock`: regenerated via `uv lock` (resolves pgvector 0.5.0).

## Testing

Built the image with the change and verified the client loads and the pgvector provider initializes:

```
from pgvector.asyncpg import register_vector  # imports cleanly
```

Confirmed against a live pgvector-backed deployment (Postgres `vector` extension 0.8.6): vector stores create/attach/query without the ModuleNotFoundError.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Added pgvector support to the project’s PostgreSQL-related dependencies.


tisnik (approved):
LGTM

PR #2719, opened 2026-09-16
RHOKP-1758: add MCP transport for OKP RAG enrichment

mwcz:
## Summary

RHOKP RAG context used to enrich LCORE responses currently comes from direct calls to the Solr API hosted in RHOKP.

This PR adds conditional support for fetching that same context via RHOKP's upcoming MCP server.  The support is conditional in the following sense: when LCORE is configured to fetch context from RHOKP, it will attempt MCP, and fall back to Solr if MCP is not available.  In this way, LCORE has support for RHOKP RAG regardless of the version of RHOKP currently deployed.  The conditional support also allows existing LCORE configuration to work with RHOKP before and after the release of the MCP server.

The plan discussed with the LCORE team is that the Solr provider (in lightspeed-providers) will be deprecated and eventually removed.  The migration to Pydantic AI will not include Solr, so the migration timeline is, in a way, the same as the timeline for Solr provider removal.

## Transport selection

- **Launch:** always wire the Solr `vector_io` provider.
- **Query:** prefer the MCP transport whenever `okp_mcp_available()` is `True`, falling back to Solr when MCP is unavailable or hard-fails for the request.  This provides compatibility with RHOKP RAG 
- **Fail-forward:** the probe result is TTL-cached (not sticky), so an upgraded RHOKP is adopted without restarting LCORE.

Product/version filtering is now **query-time only** via the request `okp` filter; the launch-time `product`/`product_version` config is dropped. `OkpMcpConfiguration` and `RH_SERVER_OKP_MCP_DEFAULT_URL` are removed; the OpenAPI schema is regenerated.

## Testing

- `uv run make format`, `black`, `pylint` (10.00/10), `pyright src` (0 errors): clean.
- `uv run make test-unit`: 3600 passed, 1 skipped.
- `make verify` still fails on `check-types-tests` (mypy-on-tests): 41 pre-existing errors across 13 files not touched here (identical at the branch base).

The RHOKP team can provide a container image of RHOKP with RAG and MCP built in (but with a small sample of the full corpus to save on build & transfer time), and we will also work on updating the internal instance of RHOKP to include MCP as well, for even easier testing.

🤖 Assisted by [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Added an optional `okp` filter for selecting RAG results by exact product and version.
  - Added RHOKP MCP support for OKP retrieval, with automatic fallback to the existing Solr-based retrieval when unavailable.
  - Applied OKP filtering consistently across query, streaming, and responses requests.

- **Documentation**
  - Documented the new request syntax, filtering behavior, transport selection, and result handling.
  - Added guidance for using the backend-neutral `okp` filter during the Solr-to-MCP transition.


docs/devel_doc/openapi.json
```diff
@@ -16255,6 +16255,93 @@
...
+            },
+            "OkpProductFilter": {
+                "properties": {
+                    "product": {
+                        "type": "string",
+                        "title": "Product",
+                        "description": "Exact product identifier (exact match, no wildcards).",
+                        "examples": [
+                            "openshift_container_platform",
+                            "rhel"
```
mwcz: `rhel` is not correct.

anik120 (commented):
This is a big PR. Are we planning on breaking it down to smaller pieces (with this PR serving as the "demo") or have we been merging big PRs like this typically?

